PR #43, opened 2026-04-21
v0.13.0: schema v7, daemon/brain handshake, installer rewrite

stevencarpenter:
Coordinated upgrade to **v0.13.0** with schema v7 migration, a daemon↔brain version handshake to prevent mid-upgrade mismatches, an installer rewrite with receipt-based idempotency, two transitive-dep CVE patches, and CLI ergonomics polish.

## Database schema (v6 → v7)

- `events` table gains `source_kind TEXT NOT NULL DEFAULT 'shell'` and `tool_name TEXT`, with a partial index on `source_kind WHERE source_kind != 'shell'`.
- `ShellEvent` struct gains `tool_name: Option<String>` (serde `#[serde(default)]` — older JSON envelopes decode as `None`).
- Producer: `claude_session.rs` now populates `tool_name = Some(pending.name.clone())` when synthesizing events from Claude Code sessions; native shell events stay `None`.
- Migration is auto-applied on startup. New dedicated integration test `schema_v7_migration.rs` covers: forward migration, existing rows default to `'shell'`, fresh DB lands at v7, and the partial index contains zero pre-migration rows.

Consumer side (brain-side policy that reads `source_kind` / `tool_name`) is **deferred** — this PR is wire-level groundwork only.

## Schema-version handshake (prevents today's 5600-error incident)

The incident that drove this: a rebuilt daemon migrated the shared DB to v7 while the installed brain was still on v6, causing every brain query to raise `RuntimeError: DB schema version mismatch`.

- Rust `storage::EXPECTED_VERSION` lifted to module-level `pub const`.
- New `brain/src/hippo_brain/schema_version.py`: single source of truth on the Python side (`EXPECTED_SCHEMA_VERSION`, `ACCEPTED_READ_VERSIONS`). `server.py::_get_conn` uses it.
- Brain `/health` now advertises `expected_schema_version` and `accepted_read_versions`.
- New `crates/hippo-daemon/src/schema_handshake.rs`: on daemon startup, probes brain's `/health` with short timeouts before `open_db`. Outcomes:
  - **Compatible** → proceed.
  - **BrainAbsent** (connection refused/timeout) → proceed with info log; brain's own guard is the backstop when it eventually starts.
  - **Unknown** (non-2xx or no version field) → warn and proceed.
  - **Incompatible** (version mismatch) → bail with a remediation message pointing at `mise run install`.

## Installer rewrite (`scripts/install.sh`)

- **Receipt system**: after successful install, writes artifact SHA-256 to `${XDG_STATE_HOME:-~/.local/state}/hippo/install-receipts/{daemon,brain,gui}.sha256`. Re-runs skip components whose receipt + install target both match. `HIPPO_FORCE=1|true|yes` bypasses.
- **Atomic writes**: `write_receipt` uses `mktemp` + same-directory rename.
- **Stage-then-swap** for brain: extract tarball to `${BRAIN_DIR}.new` with `--strip-components=1`, validate required subdirs (both `scripts/` and `shell/`), then `rm -rf old && mv new old`. Shrinks the "nothing installed" window to a single `rename(2)`.
- **GUI swap from `$TMPDIR`** rather than inside `/Applications` — outside Launch Services' scope entirely. `$TMPDIR` and `/Applications` share a device on standard APFS installs, so `mv` is an atomic rename.
- **One-time legacy cleanup**: removes the pre-v0.13 `~/.local/share/scripts/` tree (now moved to `~/.local/share/hippo-brain/scripts/`), gated on presence of a hippo-owned `hippo-*.py` file.
- **Full `--help` / usage block** documenting `HIPPO_FORCE` and install locations.

## Layout change: scripts + shell inside `brain/`

- `.github/workflows/release.yml`: the brain tarball now packs everything under a single top-level `brain/` directory; `scripts/` and `shell/` are sibling subdirs inside it.
- `crates/hippo-daemon/src/install.rs::detect_vars`: `scripts_dir = brain_dir.join("scripts")` (was `brain_dir.parent().join("scripts")`). Test pins the new path contract.
- Migration for existing installs is automatic via `hippo daemon install --force` re-rendering the xcode-ingest LaunchAgent plists with the new `__SCRIPTS_DIR__`.

## `mise run install` now syncs the brain tree

Previously rebuilt only the repo-local venv; the LaunchAgent's deployed tree at `~/.local/share/hippo-brain/` stayed stale, which was the proximate cause of the schema mismatch incident. Fix: `rsync` the repo's `brain/` into the install location and run `uv sync` there before calling `hippo daemon install`.

## `drain_brain` PID lookup

Replaces `pgrep -f "hippo-brain serve"` with parsing the PID from `launchctl list com.hippo.brain`. `pgrep -f` could match the `uv run` wrapper, its python child, and unrelated processes with the same substring; launchctl returns the single authoritative PID launchd is managing. SIGTERM to the `uv run` wrapper propagates to the child via `uv`. Added unit tests for the parser.

## CLI `--help` polish

- `hippo-brain`, `hippo-mcp`, `hippo-bulk-enrich.py`, `hippo-re-embed.py` migrated to `argparse` (previously had ad-hoc `sys.argv` parsing or none at all).
- `hippo-bulk-enrich.py` also fixes a dangling `claim_pending_events_by_cwd` import (function was removed from `hippo_brain.enrichment`; session-based claiming is the sole code path now). `--by-session` flag dropped since its alternate branch is gone.
- `scripts/hippo-monitor.sh`, `scripts/security/check-shell-secrets.sh`, `hippo-gui/scripts/{run,build}-native-app.sh` gained `usage()` functions + `-h/--help` handlers.

## CVE patches

Two transitive Python deps bumped in isolation via `uv lock --upgrade-package`:
- `python-multipart` 0.0.22 → 0.0.26 (CVSS 5.3 — DoS via oversized multipart preamble/epilogue)
- `pygments` 2.19.2 → 2.20.0 (CVSS 3.3 — ReDoS in GUID regex)

## Versioning

- `Cargo.toml` workspace version, `brain/pyproject.toml`, and `brain/uv.lock` all bumped 0.12.0 → 0.13.0.
- `docs/RELEASE.md` updated: documents the lockstep-release discipline (bump both manifests; tag `main` after merge, not the feature branch).
- `docs/FEATURE_TIMELINE.md` gains a v0.13.0 entry.

## Testing

| Check | Result |
|---|---|
| `cargo test -p hippo-core -p hippo-daemon --all-targets` | 218 pass (incl. 3 new `schema_v7_migration` + 2 new `schema_handshake` + 3 new `parse_launchctl_pid` tests) |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo fmt --check` | clean |
| `uv run pytest brain/tests` | 582 pass, 1 xfailed (pre-existing) |
| `ruff check` + `ruff format --check` | clean |
| `mise run gui:test` | 82 Swift tests pass |
| Live `hippo doctor` post-deploy | all green |

## Post-merge

Tag `main` after merge, not the feature branch:

```bash
git checkout main && git pull
git tag v0.13.0
git push origin v0.13.0
```

The release workflow builds from whatever the tag points at; tagging `main` after a squash/rebase merge avoids artifact drift.


stevencarpenter:
### Code review

PR #45, opened 2026-04-21
fix(doctor): derive session hook path at runtime, drop misleading install step

stevencarpenter:
## Summary

- `build.rs`: remove compile-time `HIPPO_SESSION_HOOK_PATH` env — released binaries were embedding the CI runner path (`/Users/runner/work/...`), causing every end-user install to show a false hook mismatch
- `commands.rs`: `check_claude_session_hook` now computes the expected path at runtime as `config.storage.data_dir.parent() + hippo-brain/shell/claude-session-hook.sh` (resolves to `~/.local/share/hippo-brain/shell/claude-session-hook.sh` on a standard install)
- `install.sh`: drop "Start services: hippo daemon start" from next steps — the installer already starts both services, so the step produced confusing "already loaded" errors

## Test Plan

- [ ] `hippo doctor` on a fresh install no longer reports hook path mismatch
- [ ] `hippo doctor` on a dev checkout (hook configured to repo path) still reports mismatch with the correct expected path
- [ ] All 215 Rust tests pass (`cargo test -p hippo-daemon -p hippo-core`)
- [ ] Fresh install next steps no longer include the redundant "start services" step

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["crates/hippo-core", "crates/hippo-daemon"]
 resolver = "2"
 
 [workspace.package]
-version = "0.13.1"
+version = "0.13.2"
 edition = "2024"
 license = "MIT"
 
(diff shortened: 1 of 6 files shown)


stevencarpenter:
@copilot resolve the merge conflicts in this pull request

PR #46, opened 2026-04-21
Derive session hook path from data_dir at runtime

stevencarpenter:
This pull request updates the way the Claude session hook path is determined and checked in the `hippo-daemon` project. Instead of embedding the expected session hook path at build time, the code now computes it dynamically at runtime based on the configured data directory. This change improves flexibility and robustness, especially across different installation locations. Additionally, the installation instructions have been updated to clarify the recommended verification step.

Session hook path computation and verification:

* The expected Claude session hook path is no longer set as a build-time environment variable in `build.rs`; instead, it is computed at runtime as one level above the Hippo data directory, making the check more flexible and robust. [[1]](diffhunk://#diff-a8129c48e7c5b92d6c38f4ba1e15efd8ac12af84851d2403ea54cddaf5351bf5L10-L22) [[2]](diffhunk://#diff-55b5ee6efd96e52056e164785025e9e6a3d75edef8ecdacdc9c015b691fe9e2aL664-R673)
* The `check_claude_session_hook` function now takes a `HippoConfig` parameter and determines the expected hook path from the configuration, rather than using a hardcoded environment variable. [[1]](diffhunk://#diff-55b5ee6efd96e52056e164785025e9e6a3d75edef8ecdacdc9c015b691fe9e2aL607-R607) [[2]](diffhunk://#diff-55b5ee6efd96e52056e164785025e9e6a3d75edef8ecdacdc9c015b691fe9e2aL664-R673)
* Output messages for session hook verification now display paths using `.display()` for better formatting, and the logic for matching configured and expected paths has been updated for accuracy.

Installation instructions:

* The installation script (`install.sh`) has been updated to recommend running `hippo doctor` immediately after configuration, removing the explicit "start services" step for clarity.

### DIFF
--- a/crates/hippo-daemon/build.rs
+++ b/crates/hippo-daemon/build.rs
@@ -7,19 +7,6 @@ fn main() {
     let version = git_describe_version(&base);
     println!("cargo:rustc-env=HIPPO_VERSION_FULL={version}");
 
-    // Embed the expected session hook path so `hippo doctor` can verify it
-    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
-    let repo_root = std::path::Path::new(&manifest_dir)
-        .parent()
-        .unwrap()
-        .parent()
-        .unwrap();
-    let hook_path = repo_root.join("shell/claude-session-hook.sh");
-    println!(
-        "cargo:rustc-env=HIPPO_SESSION_HOOK_PATH={}",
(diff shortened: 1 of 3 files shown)


stevencarpenter:
@copilot apply changes based on the comments in [this thread](https://github.com/stevencarpenter/hippo/pull/46#pullrequestreview-4146995045)

PR #54, opened 2026-04-22
fix(extension): ensure dist/ is built on install (regression from 3898910)

stevencarpenter:
## Summary

Fixes a **21-day silent outage** of the Firefox browser source. Since commit `3898910` (2026-04-01, "feat(extension): complete TypeScript rewrite"), the extension's `manifest.json` has pointed at `dist/background.js` and `dist/content.js`, but `extension/firefox/dist/` is gitignored and there was no install-pipeline step that reliably produced it. Result: the extension loaded cleanly via `about:debugging` but captured nothing. Zero browser events since 2026-04-01 23:25.

### Root cause

- `manifest.json` references `dist/background.js` + `dist/content.js` (added in 3898910).
- `extension/firefox/.gitignore` excludes `dist/`.
- The only task that built `dist/` was `build:ext`, chained under `install:ext`.
- `install:ext` short-circuits when Firefox Developer Edition isn't found, and `build:ext` silently exits 0 when `npm` is missing — so `dist/` could end up absent without any warning.
- Documented setup (CLAUDE.md) instructs users to load `extension/firefox/manifest.json` as a temporary add-on, not the side-loaded `.xpi` — and that path requires `dist/` on disk.

### Changes

- **`mise.toml`**: New `build:ext:dist` task dedicated to producing `dist/*.js`. Fails loudly (exit 1 with actionable message) when `npm` is missing, instead of silently exiting 0. `build:ext` now depends on `build:ext:dist`, and `mise run install` calls `build:ext:dist` directly as step 8b — so the dist bundle is produced whether or not Firefox side-loading succeeds.
- **`crates/hippo-daemon/src/commands.rs`**: `hippo doctor` now reports the Firefox Native Messaging manifest status and whether `extension/firefox/dist/background.js` + `dist/content.js` exist. The repo location is derived from the running binary's canonical path, so release installs (where the binary lives outside the source tree) correctly skip the dist check rather than false-alarm.

Approach **A** (build step) chosen over committing artifacts because the Node toolchain is already required for the existing `build:ext` / `install:ext` / `package:ext` flows.

## Test plan

- [x] `cargo build` — passes
- [x] `cargo clippy --all-targets -- -D warnings` — passes
- [x] `cargo fmt --check` — passes
- [x] `cargo test -p hippo-daemon --lib` — 82 tests pass (includes 3 new tests for `check_firefox_extension_dist_at`)
- [x] `mise tasks` — new `build:ext:dist` listed; syntax valid
- [x] `mise run build:ext:dist` — produces `extension/firefox/dist/background.js`, `dist/content.js`, `dist/popup.js`
- [x] `cargo run --bin hippo -- doctor` — prints `[OK] Firefox Native Messaging manifest installed` and `[OK] Firefox extension dist/ built`
- [x] Negative case: with `dist/` removed, doctor prints `[!!] Firefox extension dist/ missing: dist/background.js, dist/content.js` + `Fix: mise run build:ext:dist`

## Follow-ups

- The already-deployed `.xpi` in the user's Firefox profile is now stale (built from pre-rewrite JS). After merging, running `mise run install` will rebuild and redeploy, but users should confirm the extension is actively loaded — if they previously loaded the repo `manifest.json` as a temporary add-on, it will start working as soon as `dist/` is rebuilt (no Firefox reload required? yes reload; `about:debugging` → Reload).
- `build:ext` still exits 0 when npm is missing (preserving the "skip .xpi packaging" behavior). Only the new `build:ext:dist` fails hard. If we decide the .xpi flow is dead in favor of `about:debugging`, we can simplify further.
- Consider making `install:ext` optional (gated by a config flag) since `about:debugging` is the documented path.

Root-cause commit: `3898910`

### DIFF
--- a/crates/hippo-daemon/src/commands.rs
+++ b/crates/hippo-daemon/src/commands.rs
@@ -606,6 +606,9 @@ pub async fn handle_doctor(config: &HippoConfig) -> Result<()> {
     // Check Claude session hook
     check_claude_session_hook(config);
 
+    // Check Firefox extension build + Native Messaging manifest
+    check_firefox_extension();
+
     // Check OpenTelemetry configuration
     check_otel_status(config, &client).await;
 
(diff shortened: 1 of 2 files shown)


stevencarpenter:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/stevencarpenter/hippo/pull/54#pullrequestreview-4152531453)

PR #55, opened 2026-04-22
fix(session-hook): restore tmux -t target (regression from #48)

stevencarpenter:
### Symptom

Commit [`1330113`](https://github.com/stevencarpenter/hippo/commit/1330113) (merged via #48 on 2026-04-21 23:41 UTC) removed the `-t &#34;$TMUX_TARGET_SESSION&#34;` flag from `tmux new-window` in `shell/claude-session-hook.sh`. The commit comment claimed *&#34;Omitting -t lets tmux pick the next free index&#34;* — **this is only true when the invoking process is inside a tmux pane** (i.e. when `$TMUX` is set in the environment).

The Claude Code `SessionStart` hook runs as a direct child of Claude Code, not of tmux, so `$TMUX` is **not** inherited. Without `-t`, tmux has no current session to target, and `tmux new-window -d` either fails silently or creates a window in the wrong server/session. The hook still logs `spawned tmux window in session=` because the log line runs unconditionally after the tmux call (no exit-code check), so the regression was invisible in the debug log.

**Impact:** zero `claude_sessions` rows written since 2026-04-21 23:41 UTC. Every Claude session — mine and anyone on `main` — has been dropping.

### Fix

Use `-t &#34;${TMUX_TARGET_SESSION}:&#34;` (trailing colon, empty window name). Per `tmux(1)`:

&gt; An empty window name specifies the next unused index if appropriate (for example the `new-window` and `link-window` commands)

This restores session targeting **and** avoids the original `create window failed: index N in use` bug that motivated #48. That bug was caused by the bare `-t session_name` form: when the session name has no colon, tmux parses it as a `target-window`, and with `base-index=1` + window 1 occupied the command collided at index 1. The trailing-colon form sidesteps that parsing entirely: tmux sees session + &#34;next unused index&#34; and auto-picks.

Same fix applied to the fallback `tmux has-session -t hippo` branch (`-t hippo` → `-t &#34;hippo:&#34;`).

### How this avoids the original &#34;index N in use&#34; issue

Reproduced locally on the user&#39;s live tmux session (name `1`, `base-index 1`, window 1 occupied):

```
$ tmux new-window -d -t &#34;1&#34; -n test &#34;sleep 1&#34;
create window failed: index 1 in use          ← original #48 bug

$ tmux new-window -d -t &#34;1:&#34; -n test &#34;sleep 1&#34;
(succeeds, picks index 5)                      ← this PR
```

### Tests

New `tests/shell/test-claude-session-hook.sh` covers both failure modes explicitly:

1. **Pre-#48 regression guard** — creates a session with `base-index=1` and an occupied window 1, runs the exact `tmux new-window -d -t &#34;${TMUX_TARGET_SESSION}:&#34; -n ... cmd` invocation, asserts no `&#34;index N in use&#34;` stderr and a new window appears at a free index &gt; base-index.
2. **End-to-end hook invocation** — spawns an isolated tmux server on a private socket, stubs `hippo` on PATH, feeds representative `SessionStart` JSON on stdin, and asserts (a) the hippo window appears in the target session, (b) the window name contains the expected short-id prefix, (c) the debug log recorded the `-t session:` branch, and (d) no `&#34;index N in use&#34;` error appeared.

### Verification

All checks run locally on the worktree before pushing:

- `cargo build` — pass
- `cargo clippy --all-targets -- -D warnings` — pass
- `cargo fmt --check` — pass
- `cargo test -p hippo-daemon` — pass (all existing tests)
- `shellcheck shell/claude-session-hook.sh tests/shell/test-claude-session-hook.sh` — pass
- `bash tests/shell/test-claude-session-hook.sh` — 7/7 pass

### Follow-ups surfaced

- The hook logs &#34;spawned tmux window&#34; unconditionally regardless of whether `tmux new-window` succeeded. Worth adding `|| log &#34;WARN: tmux new-window failed with $?&#34;` so a future regression of this class doesn&#39;t go silent for another ~12 hours.
- `hippo doctor` does not currently detect &#34;hook is configured and fires, but no `claude_sessions` rows are being produced.&#34; A staleness check against the most recent `claude_sessions.created_at_ms` (e.g. warn if &gt;24h despite a live daemon) would have surfaced this sev1 within hours.

### DIFF
--- a/shell/claude-session-hook.sh
+++ b/shell/claude-session-hook.sh
@@ -88,16 +88,21 @@ TMUX_CMD="HIPPO_WATCH_PID=${CLAUDE_PID} $(printf '%q' "$HIPPO_BIN") ingest claud
 # tmux new-window -d returns immediately — the tail loop runs inside the new window,
 # so this hook never blocks Claude Code from launching.
 if [ -n "$TMUX_TARGET_SESSION" ]; then
-    # We are inside tmux: create a new window in the current session. Omitting -t
-    # lets tmux pick the next free index rather than inserting relative to a specific
-    # window, which avoids "index N in use" errors with non-default base-index configs.
-    tmux new-window -d -n "$WINDOW_NAME" "$TMUX_CMD"
+    # Target Claude's tmux session. Trailing colon (`session:`) is the key: per
+    # tmux(1), "An empty window name specifies the next unused index if
+    # appropriate (for example the new-window ... commands)". A bare
+    # `-t session_name` would be parsed as a target-window and collide with
+    # the base-index (e.g. `index 1 in use` when base-index=1 and window 1
+    # exists), which is the regression fixed here. Omitting `-t` entirely is
+    # also wrong: this hook is a child of Claude Code, not of tmux, so $TMUX
+    # may not be set and tmux would have no current session to target.
+    tmux new-window -d -t "${TMUX_TARGET_SESSION}:" -n "$WINDOW_NAME" "$TMUX_CMD"
     log "spawned tmux window in session=$TMUX_TARGET_SESSION"
 elif tmux list-sessions &>/dev/null; then
     # Not inside tmux but a tmux server is running — reuse the hippo session
     # if it already exists (from a prior fallback spawn), otherwise create it.
     if tmux has-session -t hippo 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


stevencarpenter:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/stevencarpenter/hippo/pull/55#pullrequestreview-4152541692)

PR #94, opened 2026-04-26
fix(install): false-positive shell warning + post-install verification (v0.17.1)

stevencarpenter:
## Summary

Two changes to `scripts/install.sh`, plus a patch version bump to **v0.17.1**.

### 1. Fix false-positive "missing file" warning

`warn_on_stale_shell_hook_sources` was reporting

```
[WARNING] Shell config .../personal-shell-functions.zsh sources missing file: $HOME/.local/share/hippo-brain/shell/hippo-env.zsh
[WARNING] Shell config .../personal-shell-functions.zsh sources missing file: $HOME/.local/share/hippo-brain/shell/hippo.zsh
```

every single install — even when the files were present and the hook was working — for users who source hippo via `source "$HOME/.local/share/hippo-brain/shell/..."`.

**Root cause:** the script extracts the source-path string from the user's zsh config with `sed`, then runs `[ -e "$source_path" ]`. Bash never re-evaluates `$HOME` inside the captured string, so the existence check tested the literal path `$HOME/...` and always failed.

**Fix:** added `expand_shell_path` that expands `$HOME`, `${HOME}`, and a leading `~/` before both the existence check and the case-match against the expected canonical paths. Lines that still contain unresolved `$VAR` after expansion are skipped (rather than misreported), since we can't safely `eval` arbitrary content from a user's shell config.

### 2. Post-install verification step

The installer used to print "Verify installation: hippo doctor" as a manual next step. New `verify_installation` runs that automatically so a fresh install fails loud rather than leaving the user to discover broken state later:

- Polls the daemon socket and brain `/health` for up to 60 s — brain is the slow one on cold start because of torch import. Port comes from `[brain].port` in `config.toml`, default 9175.
- Runs `hippo doctor` and prints the full output.
- If doctor reports failures, prints a hint block listing common causes (brain warming up, LM Studio not running, Firefox extension not loaded, stale shell config).
- Always returns 0, so a partial-but-recoverable install still surfaces the "Next steps" message and the user can rerun `hippo doctor`.

### 3. Version bump

- `Cargo.toml` workspace, `Cargo.lock` (hippo-core, hippo-daemon)
- `brain/pyproject.toml`

Tagging `v0.17.1` after merge will produce a fresh `install.sh` release asset that no longer carries the false-positive warning.

## Test plan

- [x] `bash -n scripts/install.sh` — syntax clean
- [x] Local end-to-end run on the affected machine: previous double WARNING is gone, brain HTTP wait succeeded, `hippo doctor` ran inline, hint block triggered correctly on the pre-existing `[!!] browser events` failure
- [x] `cargo build` — workspace builds at 0.17.1
- [ ] On merge, tag `v0.17.1` and confirm the release `install.sh` no longer warns on a clean re-install


stevencarpenter:
Addressed all three review findings in `aedcc4c`:

1. **`$HOME` substring leak** (Copilot, line 542) — Fixed. The naive `${p//\$HOME/${HOME}}` was a global substring replace, so `$HOME_DIR/foo` would have been rewritten to `<value>_DIR/foo`, eating the `$` that the `*'$'*` skip arm relies on downstream. Now replaces only slash-bounded `$HOME/` and trailing `$HOME`, leaving `$HOMEDIR`, `$HOMER_DIR`, `$XDG_DATA_HOME` etc. untouched. The `${HOME}` form is still safe to substring-replace because `}` is itself the token terminator. While in there, also fixed the `~/` strip to use `${p:2}` since bash doesn't reliably strip via `${p#~/}` (case arm already guarantees the first two chars).

2. **Hardcoded socket path** (Copilot + Codex, line 438/465) — Fixed. Dropped the `[ -S "${DATA_DIR}/daemon.sock" ]` gate. `hippo status` already resolves the socket path via `config.socket_path()` internally (including the long-path /tmp fallback and any `[storage].data_dir` override), so its exit code is the right signal.

3. **Iteration count vs wall-clock** (Copilot, line 462) — Fixed. Loop now uses bash `$SECONDS` to track elapsed wall-clock instead of iteration count. Previously worst-case was ~3× the 60s budget once `curl --max-time 2` + `sleep 1` + `hippo status` latency stacked up.

Verified `expand_shell_path` against ten edge cases including the regression cases the reviewer flagged (`$HOME_DIR`, `$HOMEDIR`, `$XDG_DATA_HOME`) — all produce the expected output. Re-ran the installer end-to-end on the affected machine: doctor reports all checks pass, no false-positive shell warnings, brain readiness detected within budget.

PR #100, opened 2026-04-27
fix(brain): enrichment quality — worktree pollution, verbatim facts, design_decisions (#98)

### LINKED ISSUE #98: Enrichment quality: worktree-path pollution + lossy summaries undermine MCP recall
## Background

While dogfooding hippo's MCP after PR #43 (`v0.13.0: schema v7, daemon/brain handshake, installer rewrite`) merged, I evaluated whether hippo had captured useful, accurate memory of the session — the kind a future agent would query via the MCP. The session was substantial (6 commits, ~1,800 LOC across 52 files, several design pivots) so it's a good stress test of the enrichment pipeline.

The TL;DR: hippo is **adequate for narrative recall** (what happened, why) but **unreliable for precision recall** (specific names, versions, symbols). And the entity graph is being **polluted by parallel-agent worktrees** in a way that compounds with every multi-agent run.

This issue captures the full investigation, four distinct findings ranked by severity, and concrete recommendations. **Worktree pollution is the most urgent** because parallel-agent worktrees fire constantly — every `Task`/`TeamCreate` invocation that sets `isolation: \"worktree\"` creates a new `.claude/worktrees/agent-<sha>/` tree. Ad-hoc auditing my own session entities suggests every commonly-edited file in the repo has 3–10 duplicate entity rows pointing at agent-worktree paths that no longer exist.

## Methodology

I ran the following queries via the hippo MCP and qualitatively scored each:

| Query | Type | Tool |
|---|---|---|
| \"What was hippo PR #43 about?\" | high-level recall | `mcp__hippo__ask` |
| \"Why did we add the schema-version handshake?\" | causal/reasoning | `mcp__hippo__ask` |
| \"How does install.sh receipt system work? What env var bypasses skip?\" | architecture + specific fact | `mcp__hippo__ask` |
| \"Why move scripts/ and shell/ inside brain/?\" | design rationale | `mcp__hippo__ask` |
| \"What CVE-bumps shipped in v0.13.0 and to what versions?\" | precise facts | `mcp__hippo__ask` |
| \"Why stage GUI from \$TMPDIR vs /Applications?\" | design pivot recall | `mcp__hippo__ask` |
| \"drain_brain pgrep launchctl uv run wrapper\" | symbol/code search | `mcp__hippo__search_knowledge` |
| `type=file query=schema_handshake` | entity coverage | `mcp__hippo__get_entities` |

Detailed responses captured in conversation transcript.

### Finding 1 — Worktree paths create duplicate entities (URGENT)

`get_entities(type=file, query=\"schema_handshake\")` returned 4 distinct file entities for what is effectively one logical source file:

```
crates/hippo-daemon/src/schema_handshake.rs                                       (canonical)
/Users/carpenter/projects/hippo/.claude/worktrees/agent-ac83d4d3/crates/hippo-daemon/src/schema_handshake.rs
/Users/carpenter/projects/hippo/.claude/worktrees/agent-a5642c4b/crates/hippo-daemon/src/schema_handshake.rs
/Users/carpenter/projects/hippo/.claude/worktrees/agent-a2e7bbb5/tests/unit/test_schema_handshake.py    (legitimately different — a test file)
```

Three of the four are duplicates created when parallel agents (Task tool with `isolation: \"worktree\"`, or TeamCreate) edited the same source from inside their isolated worktrees. The agent-XXX directories are ephemeral — they're cleaned up after the agent's work is merged or discarded — so over time these entity rows point at paths that no longer exist on disk.

**Why this matters:**
1. Entity counts in `get_entities` are inflated by an order of magnitude for any actively-edited file.
2. The knowledge graph's first_seen/last_seen for a file is fragmented across N rows instead of consolidated on one canonical entity.
3. A future agent asking \"how often is `schema_handshake.rs` touched?\" gets a misleading distribution — most of the recency comes from worktree paths that vanished.
4. Compounds linearly with parallel-agent usage. The user runs Task/TeamCreate frequently; this pollution accrues without bound.

**Forward fix (path canonicalization at enrichment time):**

When the enrichment pipeline records a file entity from a Claude session or shell event, normalize the path:
```python
def canonicalize_path(raw: Path) -> Path:
    # Strip ANY .claude/worktrees/agent-*/ prefix back to the repo root.
    parts = raw.parts
    for i, p in enumerate(parts):
        if p == \".claude\" and i + 2 < len(parts) and parts[i+1] == \"worktrees\" and parts[i+2].startswith(\"agent-\"):
            return Path(*parts[:i], *parts[i+3:])
    return raw
```

Apply this in `brain/src/hippo_brain/entity_resolver.py` (or wherever file entities are extracted from enrichment output) before the `entities` upsert.

**Backfill (reconciling existing polluted data):**

Worktree-path entities ALREADY in the database need to be merged into their canonical form. A new `brain/scripts/dedup-worktree-entities.py` migration that:
1. SELECT all `entities` rows where `name` matches `.*\\.claude/worktrees/agent-[^/]+/.*`.
2. For each, compute the canonical path (strip the worktree prefix).
3. UPSERT the canonical entity (carrying first_seen = MIN, last_seen = MAX from all duplicates).
4. UPDATE all `knowledge_node_entities` rows pointing at the old IDs to point at the canonical ID.
5. DELETE the now-orphaned worktree-path rows.

Pattern matches existing `brain/scripts/dedup-entities.py` from the v5→v6 migration. Ship as a one-shot script the user runs once, OR auto-run on next migration.

The same logic applies to **events** — events captured from a worktree should also be normalized so `cwd`/`git_repo` reflect the parent repo, not the worktree subpath.

---

### Finding 2 — Verbatim facts (versions, env var names, package names) get summarized away

Two specific demonstrations:

**A. CVE bump versions are lost.** Query: \"What CVE-bumps shipped in v0.13.0 and to what versions?\" Hippo answered: *\"not specified in the provided logs.\"* The actual commit message had `python-multipart 0.0.22 → 0.0.26` and `pygments 2.19.2 → 2.20.0` literally — the enrichment LLM paraphrased it to \"bumped to address Dependabot security alerts.\"

**B. Env var name was hallucinated.** Query: \"What env var bypasses the receipt-skip behavior?\" Hippo correctly admitted the name wasn't in context, then *guessed* `HIPPO_FORCE_INSTALL=1`. Actual name: `HIPPO_FORCE`. Honest uncertainty + plausible hallucination = a real risk that a downstream agent pastes the wrong name into a script.

**Recommendation:** tighten the enrichment system prompt (`brain/src/hippo_brain/enrichment.py::SYSTEM_PROMPT`) to require verbatim preservation of:
- Package names + versions (`name@version` format).
- Environment variable names (`UPPERCASE_WITH_UNDERSCORES=value`).
- Symbol names (function/class/const identifiers).
- File paths (already handled, but worth confirming).

Specifically, add a rule like *\"In `key_decisions` and `summary`, preserve any identifier matching `[A-Z][A-Z0-9_]{2,}` or any version-string `\\d+\\.\\d+\\.\\d+` exactly as it appeared in the source. Do not paraphrase.\"* Validate with a regression eval comparing the enrichment output against the source for these patterns.

---

### Finding 3 — Design pivots get flattened into prose, losing the \"considered X, abandoned for Y\" structure

Query: \"Why stage GUI from \$TMPDIR rather than inside /Applications? What was the earlier approach we abandoned?\" Hippo's answer:
> The previous implementation staged the GUI directly inside /Applications. That approach caused two problems: cross-device moves could fail silently... partial-failure safety was weak.

That's plausible-but-wrong. The actual abandoned approach was staging inside `/Applications/.hippo-staging/HippoGUI.app` (a hidden subdirectory in /Applications, not a hidden bundle directly in /Applications). The actual reason for abandoning it was **Launch Services crawls /Applications recursively**, which could register the transient bundle and create a duplicate. The reviewer for PR #43 specifically pushed back on the original approach with this reasoning, and the pivot is a load-bearing design decision.

This pattern — \"we considered X, picked Y, here's why\" — recurs constantly in design discussions and is exactly the kind of context a future agent needs to *not* re-propose the abandoned approach. Currently it gets flattened into prose summary and the *reason* for rejection is lost or paraphrased into something generic.

**Recommendation:** add a structured `design_decisions` field to the enrichment schema (sibling of `key_decisions`) that captures the full alternatives table:
```json
\"design_decisions\": [
  {
    \"considered\": \"Stage GUI inside /Applications/.hippo-staging/HippoGUI.app\",
    \"chosen\": \"Stage GUI in $TMPDIR, swap via mv to /Applications\",
    \"reason\": \"Launch Services crawls /Applications recursively and may register the transient hidden bundle, creating duplicate-bundle ambiguity\"
  }
]
```

Surface these in the RAG prompt context so `ask` answers can cite them directly when a future question hits the design space.

---

### Finding 4 — Semantic search misses specific symbols when the enrichment summary doesn't mention them

`mcp__hippo__search_knowledge(query=\"drain_brain pgrep launchctl uv run wrapper hippo\", mode=\"semantic\")` returned **zero hits**.

This is despite the session containing:
- A committed change (`b49e4a2`) replacing `pgrep -f \"hippo-brain serve\"` with `launchctl list com.hippo.brain` parsing in `crates/hippo-daemon/src/install.rs::drain_brain`.
- An entire conversation thread debugging *why* `pgrep -f` was problematic (`uv run` wrapper vs python child).
- 3 new unit tests for `parse_launchctl_pid`.

The embedding for this knowledge node clearly didn't preserve any of the keywords in the query. Likely root cause: the enrichment LLM summarized the change as \"replaced pgrep with launchctl\" without naming `drain_brain` or `parse_launchctl_pid`, so semantic search has no surface to bind to.

**Recommendation:** include in the enrichment system prompt: *\"The `embed_text` field should include all symbol names (functions, types, constants), file paths, and command names mentioned in the events. These are the high-signal terms a future search will use to find this node.\"* The `embed_text` field is what's vectorized for semantic search; today it skews toward natural-language summary. Tilting it toward identifier-density would close most of the gap.

A follow-up eval: take a sample of enriched nodes and check that every Cargo crate name, Python module, and bash function name in the underlying events appears verbatim in `embed_text`. Nodes failing this check are candidates for re-enrichment.

---

## Recommendations summary

| # | Recommendation | Priority | Effort |
|---|---|---|---|
| 1a | **Canonicalize worktree paths at enrichment time** in `entity_resolver.py` | URGENT | Small (~30 lines + tests) |
| 1b | **Backfill: dedup existing worktree-polluted entities** via a one-shot migration script | URGENT | Medium (similar to existing `dedup-entities.py`) |
| 1c | Apply the same canonicalization to **events** captured from worktree subpaths | URGENT | Small |
| 2 | Tighten enrichment system prompt to **preserve identifiers/versions/env-var names verbatim** | High | Small (prompt change + eval regression) |
| 3 | Add structured **`design_decisions`** field to enrichment schema | Medium | Medium (schema migration + prompt + RAG surface) |
| 4 | Bias **`embed_text`** toward symbol/identifier density | High | Small (prompt change) |

## Out-of-scope but worth tracking

- **Run an `hippo-eval`-style regression** on the enrichment system prompt before and after these changes, with a held-out set of \"specific fact\" questions (CVE versions, env var names, exact symbols).
- **Periodic worktree-cleanup job**: even with canonicalization at write-time, old entity rows accumulate. A `brain/scripts/prune-worktree-entities.py` run weekly via launchd could keep the graph clean.
- **MCP-side audit**: `get_entities` could refuse to return entries under `.claude/worktrees/...` even if the canonicalization isn't yet applied at write-time, as a defense-in-depth measure.

## Acceptance criteria for closing this issue

All four recommendations land. Specifically:

- [ ] Forward fix: new file/event entities created in `.claude/worktrees/agent-*/` paths are stored under their canonical repo path in the entities table.
- [ ] Backfill script runs cleanly on the live DB and reduces `SELECT count(*) FROM entities WHERE name LIKE '%.claude/worktrees/agent-%'` to zero (or near-zero — only legitimately worktree-only artifacts remain).
- [ ] Enrichment prompt updated; verifiable with a manual probe against a known recent commit (e.g., \"what version is python-multipart in v0.13.0?\" → returns `0.0.26`).
- [ ] `embed_text` density of identifiers verified against a sample.
- [ ] (Optional but recommended) `design_decisions` field shipped and visible in `mcp__hippo__ask` responses.

stevencarpenter:
Closes #98.

## Summary

Four fixes for the four findings in #98 — enrichment-quality issues that undermined MCP recall during dogfooding of v0.13.0.

### F1 — Worktree path canonicalization (URGENT)
- New `strip_worktree_prefix()` in [entity_resolver.py](brain/src/hippo_brain/entity_resolver.py) strips `.claude/worktrees/<X>/` segments from path-type entity values.
- **Bug fix vs. issue's proposed regex:** the original regex checked `parts[i+2].startswith("agent-")`, which silently misses `feat-*`, `competent-*`, `gracious-*`, and other naming schemes Claude Code uses for worktrees. The new regex (`\.claude/worktrees/[^/]+/`) is segment-name agnostic.
- Applied at canonicalize time **before** the project-root strip (otherwise the worktree path no longer matches the project-root prefix).
- Same `strip_worktree_prefix()` is applied to `cwd` in all three prompt builders (shell, Claude session, Codex session) so the LLM sees the parent-repo path. The events table itself is **not** mutated — raw cwds are preserved.
- The existing [`dedup-entities.py`](brain/scripts/dedup-entities.py) script now sweeps up worktree pollution automatically because it re-canonicalizes through `canonicalize()`.

**Live-DB validation** (dry-run): 513 worktree-polluted rows, 359 duplicate groups queued for merge.

### F2 — Verbatim preservation
All three SYSTEM_PROMPTs now require exact preservation of identifiers (`[A-Z][A-Z0-9_]{2,}`), versions (`\d+\.\d+\.\d+`), package@version pairs, symbol names, and CLI flags. Explicit instruction: **omit rather than guess** — a hallucinated `HIPPO_FORCE_INSTALL` is worse than admitting `HIPPO_FORCE` was unknown.

### F4 — Identifier-dense embed_text
Same three prompts now describe `embed_text` as a search-friendly tag soup of symbol names, file paths, and command names — not polished prose. Density of search keys beats elegance for semantic retrieval.

### F3 — Structured `design_decisions`
- New optional `design_decisions: list[dict]` on `EnrichmentResult`. Each entry has `considered`, `chosen`, `reason` (all required strings; partial entries dropped during validation).
- Added to all three prompts.
- Persisted in `knowledge_nodes.content` JSON (no schema migration).
- Extracted in [retrieval.py](brain/src/hippo_brain/retrieval.py:642) and surfaced in [rag.py:_hit_lines](brain/src/hippo_brain/rag.py:107) so synthesis can answer "why X over Y" questions accurately.

## Acceptance criteria (from #98)

- [x] Forward fix: new file/event entities created in `.claude/worktrees/agent-*/` paths are stored under their canonical repo path
- [x] Backfill: existing `dedup-entities.py` collapses worktree pollution (live dry-run: 359 merges)
- [x] Enrichment prompt updated for verbatim preservation (all 3 prompts)
- [x] `embed_text` density biased toward identifiers (all 3 prompts)
- [x] `design_decisions` field shipped end-to-end (model → prompt → writer → retrieval → RAG render)

## Test plan

- [x] `uv run --project brain pytest brain/tests` — 746 passed, 1 skipped
- [x] `cargo test --workspace` — clean (no Rust changes, sanity check)
- [x] Manual `dedup-entities.py --dry-run` on live DB
- [ ] Operator runs `uv run --project brain python brain/scripts/dedup-entities.py` (no `--dry-run`) post-merge to actually collapse the 359 duplicate groups
- [ ] Probe enrichment quality on a future session (manual: ask the MCP about a recent commit's CVE bumps and verify the version strings come back verbatim)

## Notes / risks

- `design_decisions` is **only populated for new enrichments**. Old knowledge nodes have an empty list. The `_extract_design_decisions` helper handles missing keys gracefully.
- `workflow_enrichment.py` was reviewed but **not** changed: it produces a free-text summary, not a structured `EnrichmentResult`. No design_decisions or verbatim rules apply there.

### DIFF
--- a/brain/scripts/dedup-entities.py
+++ b/brain/scripts/dedup-entities.py
@@ -1,10 +1,22 @@
 #!/usr/bin/env python3
-"""One-shot entity deduplication migration (R-16).
+"""One-shot entity deduplication migration.
 
-Re-canonicalizes all entities using the new entity_resolver rules and merges
-rows where (type, new_canonical) would collide. Keeps the oldest row (smallest
-created_at), re-points all knowledge_node_entities to the survivor, then
-deletes the duplicate entity rows.
+Re-canonicalizes all entities using the current entity_resolver rules and
+merges rows where (type, new_canonical) would collide. Keeps the oldest row
+(smallest created_at), re-points all knowledge_node_entities to the survivor,
+then deletes the duplicate entity rows.
+
+This script handles two flavors of duplication transparently because it
+delegates to canonicalize():
+
+  1. Multiple project roots resolving to the same logical file (e.g.
+     hippo vs. hippo-postgres) — the original v5→v6 use case.
+  2. Ephemeral parallel-agent worktrees under .claude/worktrees/<X>/...
+     (issue #98). Worktree subdirectory names vary (`agent-*`, `feat-*`,
+     adjective-noun-hex), and the directories are removed once the agent's
+     work is merged or discarded — leaving polluted entity rows. Re-running
+     this script after canonicalize() learns the worktree-strip rule
+     collapses N copies of every commonly-edited file to one canonical row.
 
 Usage:
     uv run --project brain python brain/scripts/dedup-entities.py [--data-dir PATH] [--dry-run]
@@ -13,7 +25,7 @@ Environment:
     HIPPO_PROJECT_ROOTS  Colon-separated list of absolute project root paths to
                          strip from path-type entity values (e.g.
                          /home/user/projects/hippo:/home/user/projects/hippo-postgres).
-                         Required for worktree-prefix dedup to work correctly.
+                         Required for project-root dedup to work correctly.
 
 Exits 0 on success. Non-zero on error.
 """
--- a/brain/src/hippo_brain/browser_enrichment.py
+++ b/brain/src/hippo_brain/browser_enrichment.py
@@ -20,13 +20,24 @@ Extract what they were researching, learning, or investigating. Focus on technic
 
 IMPORTANT: Be specific. Use actual page titles, URLs, technical concepts, and search queries from the data. Generic descriptions like "browsed some pages" are unacceptable.
 
-The embed_text field should read like a developer's research log — specific enough that searching for "Rust Display trait implementation" or "cargo proc-macro error" would find it.
+VERBATIM PRESERVATION RULE: In every text field (summary, intent, key_decisions, problems_encountered, design_decisions, embed_text), reproduce the following kinds of tokens EXACTLY as they appeared in the page titles, URLs, or search queries. Do NOT paraphrase, normalize, or guess at them:
+  - Environment variable names (UPPERCASE_WITH_UNDERSCORES)
+  - Constants and ALL_CAPS identifiers matching [A-Z][A-Z0-9_]{2,}
+  - Semantic versions matching \\d+\\.\\d+\\.\\d+ (e.g. 1.75.0, 22.3.1)
+  - Package@version pairs (e.g. "tokio@1.40.0")
+  - Symbol names: function, method, class, struct, trait, and constant identifiers
+  - CLI flag names (--release, --no-default-features)
+  - File paths and command names
+If you are unsure of an exact name or version, OMIT it rather than guess.
+
+The embed_text field is what powers semantic search over this research session. It MUST be identifier-dense: include every symbol name, package name, version string, search query, and technical term that appears in the page data. Density of identifiers beats prose elegance — a future agent will search this field by keyword. A good embed_text reads like a tag soup of the actual technical content (e.g. "tokio mpsc unbounded_channel rust async-trait stackoverflow tokio-rs docs.rs"), not like a polished paragraph.
 
 Output a JSON object with these fields:
 - summary: Specific description of what was researched or learned
 - intent: The developer's goal (e.g., "research", "debugging", "learning", "reference")
 - outcome: One of "success", "partial", "failure", "unknown"
 - key_decisions: List of decisions informed by the research
+- design_decisions: List of "considered X, chose Y, reason Z" structured decisions when the research shows an alternative was evaluated and rejected. Each entry is an object with keys "considered", "chosen", and "reason". Empty list if no alternatives were weighed.
 - problems_encountered: List of obstacles or dead ends
 - entities: An object with lists of extracted entities:
   - projects: Project names mentioned or inferred
@@ -36,7 +47,7 @@ Output a JSON object with these fields:
   - errors: Error messages being researched
   - domains: Key domains visited (e.g., "stackoverflow.com", "docs.rs")
 - tags: Descriptive, specific tags
-- embed_text: A detailed paragraph describing the research session. Specific topics, search queries, and sources. Optimized for semantic search.
+- embed_text: A detailed, identifier-dense paragraph (see rule above). Optimized for keyword retrieval, not prose.
 
 Output ONLY valid JSON, no markdown fences or extra text."""
 
@@ -303,6 +314,7 @@ def write_browser_knowledge_node(
             "tags": result.tags,
             "key_decisions": result.key_decisions,
             "problems_encountered": result.problems_encountered,
+            "design_decisions": result.design_decisions,
         }
     )
     tags_json = json.dumps(result.tags)
--- a/brain/src/hippo_brain/claude_sessions.py
+++ b/brain/src/hippo_brain/claude_sessions.py
@@ -12,6 +12,7 @@ from hippo_brain.enrichment import (
     is_enrichment_eligible,
     upsert_entities,
 )
+from hippo_brain.entity_resolver import strip_worktree_prefix
 from hippo_brain.models import EnrichmentResult
 from hippo_brain.watchdog import DEFAULT_LOCK_TIMEOUT_MS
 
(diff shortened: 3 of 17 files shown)


stevencarpenter:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/stevencarpenter/hippo/pull/100#pullrequestreview-4178889344)

stevencarpenter:
@copilot+gpt-5.4 Comprehensively review this PR. Also take into account any intersections or potential conflicts with the other open pull request against hippo.

Issue #99, opened 2026-04-27
Investigate duplicate trailing segment_index in claude_sessions


stevencarpenter:
## Symptom

`claude_sessions` rows occasionally land with two consecutive `segment_index` values (`N` and `N+1`) carrying **identical** `message_count`, `start_time`, `end_time`, and `tool_calls_json` for the same `session_id`. The pairs look like a row was written twice into adjacent indexes during a watcher reparse.

## Prevalence (this machine, 2026-04-27)

```sql
WITH dups AS (
  SELECT a.session_id, a.segment_index AS i1, b.segment_index AS i2
  FROM claude_sessions a
  JOIN claude_sessions b ON a.session_id = b.session_id
                          AND b.segment_index = a.segment_index + 1
                          AND a.message_count = b.message_count
                          AND a.start_time = b.start_time
                          AND a.end_time = b.end_time
)
SELECT COUNT(*) AS duplicate_pairs, COUNT(DISTINCT session_id) AS sessions
FROM dups;
-- → 44 duplicate_pairs, 44 sessions
```

Earliest example seen: 2026-04-08 20:10:31. Pattern spans the entire watcher era, predates the Bug A truncation regression discovered separately.

## Examples

| session_id | (i1, i2) | message_count | start_time |
|---|---|---|---|
| `9069be39-8e91-4a45-981d-61ea15d8043a` | (5, 6) | 18 | 2026-04-08 20:10:31 |
| `b289e412-9daf-49c3-b945-74366a8d91a6` | (8, 9) | 116 | 2026-04-08 22:15:05 |
| `296e9905-6ce8-415a-a016-0188693f888a` | (18, 19) | 36 | 2026-04-21 01:34:08 |

## Hypothesis to investigate

`extract_segments` in `crates/hippo-daemon/src/claude_session.rs` has a "finalize trailing segment" block (around line 785) that pushes the `current` segment if it has any content. There's also a split-on-user-gap path (line 692) that pushes the prior segment and starts a new one. A theory: under certain content shapes (very short trailing segment, or a final user-gap split happening close to file end), the same content is pushed both as the trailing segment of one parse AND as the start of a "new" segment in the next-up index, then on the next reparse the trailing-segment logic adds another duplicate at `N+1` that wasn't present in the prior parse.

Alternative theory: a race in the watcher where two FSEvents notifications process the same file concurrently and produce overlapping segment indexes.

## Investigation steps

[...]

PR #101, opened 2026-04-27
fix(capture): Phase 1 — watcher data-loss fix (Bug A: segment truncation)

stevencarpenter:
## Summary

Fixes a P0 data-loss bug in the Claude-session FS-watcher that has been silently truncating session segments since T-7 shipped (2026-04-25 evening). Root cause: `insert_segments` used `INSERT OR IGNORE` on `(session_id, segment_index)`, so the watcher's full-file-reparse pattern would write a segment once at first sight (with whatever content was in the file at that moment) and then silently drop every subsequent reparse with more content. Median segment in the post-T-7 window contained 2–4 messages out of hundreds; `tool_calls_json` was empty in 59/63 sessions over a 24-hour sample.

The fix replaces `INSERT OR IGNORE` with proper upsert plus a hash-based dedup mechanism for enrichment. Reparses now keep the DB in sync with the file content; enrichment re-fires only when content actually changes; a settling-sweep backstop catches segments where the user walked away mid-conversation.

Discovered while investigating I-3, reproduced live by resuming a 5-day-old session and watching segment 20 land with 4 messages / 0 tools while the file gained 117 lines / 13 tool_uses. See [`docs/capture-reliability/11-watcher-data-loss-fix.md`](docs/capture-reliability/11-watcher-data-loss-fix.md) for the full failure analysis, design rationale, and recovery story.

**Bug B** (no `events.source_kind='claude-tool'` rows produced by the watcher path) is intentionally **not** addressed here — it's tracked as Phase 2 in the same plan doc, deferred so this fix could ship as one coherent change.

## What's in the PR

- **T-A.1 — Schema v11→v12:** adds `claude_sessions.content_hash` and `claude_sessions.last_enriched_content_hash`. Idempotent ALTERs mirroring the v10→v11 pattern. Brain `EXPECTED_SCHEMA_VERSION` bumped, v11 retained in `ACCEPTED_READ_VERSIONS`.
- **T-A.2 — `compute_segment_content_hash`:** SHA256 of `tool_calls_json + user_prompts_json + assistant_texts.join("\n")`, lowercase hex.
- **T-A.3 — Upsert in `insert_segments`:** `INSERT … ON CONFLICT(session_id, segment_index) DO UPDATE SET (mutable cols, content_hash)`. Identifying columns (project_dir, source_file, parent_session_id, etc.) untouched on conflict; `last_enriched_content_hash` only ever written by the brain.
- **T-A.4 — `decide_enqueue` gate:** post-upsert, decide whether to enqueue. **INSERT path** always enqueues. **UPDATE path** enqueues iff `current_hash != prior_last_enriched_hash` AND queue.updated_at older than the 5-min debounce AND queue.status != 'processing'. Empty segments short-circuit (no enqueue).
- **T-A.5 — Brain hash propagation:** `claim_pending_claude_segments` SELECTs `content_hash`. On enrichment success, `UPDATE claude_sessions SET last_enriched_content_hash = ?`. Failure path does NOT write the hash (preserves prior value so dedup keeps working across retries).
- **T-A.6 — Watcher heartbeat settling sweep:** `run_settling_sweep` wired into the existing 30s heartbeat tick. Threshold: 30 min of file-mtime idle. Cap: 10 enqueues per tick. Pre-migration safe (graceful no-op when columns absent, with warn-throttle).
- **T-A.7 — Backfill CLI:** `hippo ingest claude-session-backfill <glob> [--since YYYY-MM-DD] [--dry-run]`. Resets `claude_session_offsets.size_at_last_read = 0` and reparses via the new upsert path. Idempotent under content_hash dedup — re-runs against unchanged content produce zero updates.
- **T-A.9 — Docs:** new AP-12 anti-pattern entry capturing the lesson; 09-test-matrix entries for the new tests; status block in 11-watcher-data-loss-fix updated to "shipped".

The duplicate-segment-index anomaly observed in the live DB during investigation (44 affected sessions; consecutive segment_indexes with identical content) is filed as #99 — independent failure mode, not addressed here.

## Architecture notes

- **Why hash dedup:** counter-based change detection misses content-replacement edits (e.g., redaction logic improving between enrichments). SHA256 is robust; cost is microseconds per upsert.
- **Why brain doesn't compute the hash:** to avoid cross-language hash divergence, brain just propagates the value the daemon wrote. `claim_pending_claude_segments` reads it; success path writes it back to `last_enriched_content_hash`.
- **Why 5-min debounce + 30-min sweep:** enrichment serves the *next* session, not the current one. Sub-5-min freshness has no UX value; completeness when the next session starts is the goal. The 30-min sweep is the backstop for the "user walked away" case.
- **Why two columns instead of one:** `content_hash` is a watcher-side write-on-every-upsert ledger; `last_enriched_content_hash` is a brain-side write-on-success ledger. Comparing them is the dedup signal. Folding into one column conflates ownership.

## Out of scope (deliberate)

- **Bug B** — no path writes `events.source_kind='claude-tool'` rows from the watcher. Tracked as Phase 2 in the plan doc; structurally simpler (stable `tool_use_id` → `INSERT OR IGNORE` actually correct there). Will land as a follow-up PR.
- **Live backfill on production DB** — Phase 1 ships the recovery tool; running it against `~/.local/share/hippo/hippo.db` is a separate operational step that needs the daemon redeployed first.
- **Phase 3 follow-ups** — resumed I-3 implementation, probe TTL pruning, duplicate segment_index investigation (#99), incremental parse, differential enrichment.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo test -p hippo-core -p hippo-daemon --lib` — 179 pass; 4 pre-existing socket-flakes unchanged (verification confirmed `daemon.rs` byte-identical to `c290ab2` v0.17.0)
- [x] `cargo test -p hippo-core -p hippo-daemon --tests` — 161 integration tests pass
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `uv run --project brain pytest brain/tests` — 727 passed (incl. 4 new `TestContentHashPropagation` tests)
- [x] `uv run --project brain ruff check brain/` clean
- [x] `uv run --project brain ruff format --check brain/` clean
- [x] CLI smoke: `hippo ingest claude-session-backfill --help` builds and prints
- [x] Live reproduction validated against session `296e9905` (see plan doc)

35+ new tests added across the diff:
- `test_hash_*` (5) — hash function determinism + sensitivity
- `test_upsert_*` (3) — upsert insert/update/idempotence
- `test_decide_enqueue_*` (6) — every branch of the gate
- `test_insert_segments_skips_enqueue_for_empty_segment` (1)
- `test_sweep_*` (9) — every branch of the settling sweep, incl. pre-migration
- `test_backfill_*` (6) + parsing tests (3) — CLI behavior end-to-end
- `test_migrate_v11_to_v12_*` (2) — schema migration + partial-success recovery
- `TestContentHashPropagation` (4) — brain claim/write paths in Python

## Reviewer checklist

- [ ] Read the plan doc first: [`docs/capture-reliability/11-watcher-data-loss-fix.md`](docs/capture-reliability/11-watcher-data-loss-fix.md). Sections "Live reproduction" and "Confirmed design facts" are the most load-bearing for understanding the fix shape.
- [ ] Schema migration: `add_column_if_missing` is idempotent and matches v10→v11 pattern; v11 in brain's `ACCEPTED_READ_VERSIONS` allows safe rollback.
- [ ] `decide_enqueue` rule order: `was_insert` → `processing` → `hash_eq` → `debounce`. Order matters — flipping any pair changes the dedup contract.
- [ ] Backfill CLI: `--dry-run` does no DB writes; idempotent on second run; `--since` filter uses local-time interpretation.
- [ ] Settling sweep: 10/tick cap and 30-min threshold are deliberate (see decision log in plan doc).

## Post-merge plan

1. `mise run install` to redeploy the daemon binary and restart `com.hippo.claude-session-watcher`. After this point, all *future* sessions are captured correctly.
2. Run `hippo ingest claude-session-backfill '~/.claude/projects/**/*.jsonl' --since 2026-04-25 --dry-run` to preview the recovery scope.
3. Run without `--dry-run` to actually recover. Idempotent — safe to re-run if interrupted.
4. Open Phase 2 PR for Bug B (per-tool events emission).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,6 +605,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "glob"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
+
+[[package]]
 name = "h2"
 version = "0.4.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -694,6 +700,7 @@ dependencies = [
  "chrono",
  "clap",
  "dirs",
+ "glob",
  "hippo-core",
  "hostname",
  "libc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -707,6 +707,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
+ "sha2",
  "sysinfo",
  "tempfile",
  "tokio",
--- a/brain/src/hippo_brain/claude_sessions.py
+++ b/brain/src/hippo_brain/claude_sessions.py
@@ -697,7 +697,7 @@ def write_claude_knowledge_node(
         # where the daemon has not yet written a hash) to avoid clobbering an
         # existing last_enriched_content_hash with NULL.
         if content_hashes is not None:
-            for seg_id, ch in zip(segment_ids, content_hashes):
+            for seg_id, ch in zip(segment_ids, content_hashes, strict=True):
                 if ch is not None:
                     conn.execute(
                         "UPDATE claude_sessions SET last_enriched_content_hash = ? WHERE id = ?",
--- a/brain/src/hippo_brain/claude_sessions.py
+++ b/brain/src/hippo_brain/claude_sessions.py
@@ -552,7 +552,8 @@ def claim_pending_claude_segments(
             f"""
             SELECT id, session_id, project_dir, cwd, git_branch, segment_index,
                    start_time, end_time, summary_text, tool_calls_json,
-                   user_prompts_json, message_count, token_count, is_subagent
+                   user_prompts_json, message_count, token_count, is_subagent,
+                   content_hash
             FROM claude_sessions
             WHERE id IN ({placeholders}) AND probe_tag IS NULL
             ORDER BY start_time ASC
@@ -578,6 +579,7 @@ def claim_pending_claude_segments(
                     "message_count": row[11],
                     "token_count": row[12],
                     "is_subagent": row[13],
+                    "content_hash": row[14],
                 }
             )
 
@@ -615,9 +617,22 @@ def _skip_ineligible_claude_segments(conn, segments: list[dict]) -> list[dict]:
 
 
 def write_claude_knowledge_node(
-    conn, result: EnrichmentResult, segment_ids: list[int], model_name: str
+    conn,
+    result: EnrichmentResult,
+    segment_ids: list[int],
+    model_name: str,
+    content_hashes: list[str | None] | None = None,
 ) -> int:
-    """Insert knowledge node linked to Claude session segments."""
+    """Insert knowledge node linked to Claude session segments.
+
+    ``content_hashes`` must be the same length as ``segment_ids`` when
+    provided.  Each entry is the ``content_hash`` value read from
+    ``claude_sessions`` at claim time (daemon-computed; brain never recomputes
+    it).  For each segment whose hash is not ``None``, the corresponding
+    ``last_enriched_content_hash`` column is updated so the watcher dedup
+    logic can detect future content changes.  ``None`` entries are skipped to
+    avoid overwriting an existing value with NULL on legacy rows.
+    """
     node_uuid = str(uuid.uuid4())
     now_ms = int(time.time() * 1000)
     content = json.dumps(
@@ -676,6 +691,18 @@ def write_claude_knowledge_node(
             """,
             [now_ms, *segment_ids],
         )
+
+        # Record the content version that was just enriched so the watcher
+        # dedup logic can detect future changes.  Skip NULL hashes (legacy rows
+        # where the daemon has not yet written a hash) to avoid clobbering an
+        # existing last_enriched_content_hash with NULL.
+        if content_hashes is not None:
+            for seg_id, ch in zip(segment_ids, content_hashes):
+                if ch is not None:
+                    conn.execute(
+                        "UPDATE claude_sessions SET last_enriched_content_hash = ? WHERE id = ?",
+                        (ch, seg_id),
+                    )
 
         # Upsert entities
         upsert_entities(conn, node_id, result.entities, SHELL_ENTITY_TYPE_MAP, now_ms)
--- a/brain/src/hippo_brain/schema_version.py
+++ b/brain/src/hippo_brain/schema_version.py
@@ -10,24 +10,28 @@ Keep `EXPECTED_SCHEMA_VERSION` in sync with `EXPECTED_VERSION` in
 migration, bump both together.
 
 v11→v12 adds `content_hash` and `last_enriched_content_hash` to
-`claude_sessions`. Brain reads neither column yet — that wiring lands in
-T-A.5 — so a v11-aware brain handles a v12 DB transparently.
+`claude_sessions`. Brain now both reads `content_hash`
+(`claim_pending_claude_segments`) and writes `last_enriched_content_hash`
+(`write_claude_knowledge_node`), so v12 is the minimum readable version.
+The daemon-side handshake (`schema_handshake.rs`) already enforces strict
+equality, and dropping v11 from `ACCEPTED_READ_VERSIONS` brings brain's
+DB-attach guard in line: pre-v12 DBs are rejected at connect time rather
+than crashing later inside the enrichment loop on `no such column:
+content_hash`.
 """
 
 from __future__ import annotations
 
 EXPECTED_SCHEMA_VERSION: int = 12
 
-# Versions brain can read without erroring, so the daemon can migrate
-# forward and brain can still serve queries during the window where the
-# new rows are settling in. Brain requires v5 as the minimum because the
-# knowledge_nodes table and FTS5 index were added in that migration; v1–v4
-# DBs must be migrated by the daemon before brain starts. v10 is kept for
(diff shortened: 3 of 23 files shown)


stevencarpenter:
## Review feedback addressed (commits `06106e2`, `58d21bf`)

Two fix-up commits land:

### Code (`06106e2`)
- **C-1 (correction to my own PR claim above):** The 4 `daemon::tests::test_*` socket failures were **not** pre-existing on main — they pass on main and failed on this branch only because `test_config()` defaulted `brain.port = 9175`, hitting the live local `hippo-brain` at v11 against the v12 daemon → strict schema-handshake bail. Test isolation defect, now fixed by binding an ephemeral `127.0.0.1:0` port. Apologies for the misleading claim — I was running the tests on a machine with a v11 brain and didn't recognize the schema-handshake interaction. **All 4 socket tests now pass cleanly.** Daemon lib test count: **187 passed / 0 failed** (was 179/4).
- **I-1:** `INSERT OR REPLACE` on `claude_enrichment_queue` (both call sites) → `ON CONFLICT(claude_session_id) DO UPDATE WHERE status != 'processing'`. Preserves brain's `locked_at`/`locked_by` under race; resets `retry_count` only on legitimate new-content versions; clears stale `error_message`. Defence-in-depth WHERE clause complements the existing `decide_enqueue` processing-skip rule.
- **I-3:** `insert_segments` and `run_settling_sweep` per-call work wrapped in `conn.unchecked_transaction()`. Batches fsyncs (large win for backfill across hundreds of files) and tightens the I-1 race window further.
- **I-4 (option B):** new `check_backfill_needed()` runs once at watcher startup; if any post-2026-04-25 segment has `content_hash IS NULL`, `warn!()`s with the exact `hippo ingest claude-session-backfill` command. `OnceLock` guard + pre-migration safe.
- **S-3:** brain `zip(segment_ids, content_hashes, strict=True)` for refactor safety.

New regression tests:
- `test_enqueue_does_not_clobber_processing_lock` — exercises the I-1 race scenario directly: seeds a `processing` row, runs the queue UPSERT SQL, asserts `locked_by` and `retry_count` survive.
- `test_insert_idempotence_preserves_queue_retry_count` — I-3 transactional behavior.
- `test_check_backfill_needed_warns_when_null_hash_post_cutoff` and `test_check_backfill_needed_silent_when_hash_set` — I-4 startup warn.

### Docs (`58d21bf`)
- **I-2:** 25 test-name rows in `09-test-matrix.md` reconciled to actual code names; non-existent test rows dropped; missed tests added.
- **S-1:** Plan doc hash-input description corrected: `assistant_texts.join("\n")` (not "assistant text count").
- **I-4 doc:** Plan-doc design bullet for the startup-warn helper, decision-log entry justifying option B over A, and a new "Install ordering" subsection documenting the daemon↔brain v12/v11 handshake constraint.

### Verification on the post-fix branch tip
- `cargo test -p hippo-daemon --lib`: **187 passed / 0 failed** ✅
- `cargo test -p hippo-core -p hippo-daemon --tests`: all green ✅
- `cargo clippy --all-targets -- -D warnings`: clean ✅
- `cargo fmt --check`: clean ✅
- `uv run --project brain pytest brain/tests`: **727 passed**, 1 skipped, 1 xfailed, 1 xpassed ✅
- `ruff check` / `ruff format`: clean ✅

Skipped from the review (deferred or non-blocking):
- **S-2** (`git_branch` not refreshed on conflict): defer — reviewer marked minor, no clear win in fixing without ambiguity.
- **S-4** (sweep over-fetch → OFFSET): defer until 1000+ unsettled segments is realistic.
- **S-5** (generic test-isolation pattern): folded into C-1 fix scope (`bind_local_http_listener`-style ephemeral port).

stevencarpenter:
## Round-2 review polish (commit `472883e`)

All six minor items addressed (none were blocking; reviewer signed off):

- **M-1**: `run_settling_sweep` doc-block contract now correctly describes `ON CONFLICT … WHERE status != 'processing'` (was still saying `INSERT OR REPLACE`).
- **M-2**: Sweep `enqueued` counter respects `tx.execute()` return value — no longer overcounts the rare race-window no-op.
- **M-3**: `insert_segments` now uses `tx.execute()` / `tx.query_row()` / `tx.last_insert_rowid()` consistently inside the transaction scope (matches `hippo-core/src/storage.rs:724-771` pattern).
- **M-4**: Renamed `test_insert_idempotence_preserves_queue_retry_count` → `test_process_file_short_circuit_preserves_queue_state`. Reviewer was right — the test exercises the file-size short-circuit in `process_file`, not the I-3 transaction. Doc comment now cross-references `test_upsert_idempotent_on_same_content` (in `claude_session.rs::tests`) which is the real I-3 idempotence coverage.
- **M-5**: Install-ordering doc note rewritten. Previously implied `ACCEPTED_READ_VERSIONS` loosened the daemon handshake; the reviewer correctly noted the daemon handshake at `schema_handshake.rs:96` is **strict equality** (`brain_expects == daemon_expected`), and `ACCEPTED_READ_VERSIONS` is brain's own DB-attach guard — an independent concern. Both version-skew directions now correctly noted as incompatible.
- **M-6**: F-25 prose row in `09-test-matrix.md` replaced with grepable list of the four T-A.10 test names.

Verification:
- `cargo test -p hippo-daemon --lib`: 187 passed / 0 failed ✅
- `cargo clippy --all-targets -- -D warnings`: clean ✅
- `cargo fmt --check`: clean ✅

Should be fully ready for merge now.

PR #105, opened 2026-04-28
fix(brain): strip worktree prefix from entity display name (#98 follow-up)

stevencarpenter:
Follow-up to PR #100. Reopens #98.

## Summary

PR #100 fixed the `canonical` column in `entities` so worktree-path entities dedup correctly at the unique-key level. **The `name` column was still storing the raw LLM-emitted path**, including `.claude/worktrees/<X>/...` prefixes. Since `mcp__hippo__get_entities` and any UI surface read the `name` column, the user-visible representation of files remained polluted, and the existing `ON CONFLICT` clause only updated `last_seen` — never `name` — so any file whose first-ever sighting came from inside a worktree was stuck with a worktree-display-name forever.

## Live-DB evidence (before this PR)

```
F1: worktree-polluted entity rows         | 558
F1: post-PR-100 first_seen (NEW pollution!) | 38
```

The 38 newly-polluted rows were created **after** PR #100 merged at `2026-04-27 09:35Z` — newest sighting `2026-04-27 14:39` from Ralph Loop runs `youthful-kirch-7d3b27`, `upbeat-wilson-1aeb94`, `agent-ade96bc9611939e0d`.

## What changed

`brain/src/hippo_brain/enrichment.py` — `upsert_entities` now:

1. Strips the worktree prefix from the **display name** before insert (`strip_worktree_prefix(name)`), not just the canonical.
2. On conflict, **repairs** an existing polluted name (`name LIKE '%.claude/worktrees/%'`) by overwriting it with the new clean value.
3. **Preserves** clean existing names — the `CASE WHEN` avoids churning stable rows when both writes are clean (just dedup on canonical).

```sql
ON CONFLICT (type, canonical) DO
UPDATE SET last_seen = excluded.last_seen,
           name = CASE
               WHEN name LIKE '%.claude/worktrees/%'
                   THEN excluded.name
               ELSE name
           END
```

## Tests (3 new, all passing)

- `test_upsert_entities_strips_worktree_from_display_name` — fresh write with worktree path stores the stripped form.
- `test_upsert_entities_repairs_existing_polluted_name` — clean write heals a pre-fix polluted row.
- `test_upsert_entities_does_not_churn_clean_existing_name` — first write with absolute path, second write with relative path, original wins.

## Test plan

- [x] `uv run --project brain pytest brain/tests` — **757 passed**, 1 skipped, 1 xfailed, 1 xpassed
- [ ] Operator runs `uv run --project brain python brain/scripts/dedup-entities.py` post-merge to collapse the 558 existing polluted rows (script was ready since PR #100 but never executed).
- [ ] Post-merge verification probe: `mcp__hippo__get_entities(type='file', query='enrichment.py')` should return clean paths only.

## What this PR does NOT fix

- **F2 (verbatim facts)** — empirically still broken; MCP probe returns *"version not specified"* for the canonical CVE-bump test case despite PR #100's prompt change. The model (`gpt-oss-120b`) is paraphrasing past the prompt instruction. Detailed in the reopened-#98 comment with three options for follow-up.
- **58 malformed-JSON `content` rows** — separate issue, possibly related to PR #104's `strict=False` fix.

## Notes / risks

- The `CASE WHEN` clause means existing **clean** rows that happen to share a canonical with a new write keep their original display name. This is intentional (avoids churn) but means we won't, e.g., normalize `/Users/foo/...` → `~/foo/...` over time. Acceptable tradeoff.
- The fix is forward-only — existing 558 polluted rows still need the dedup-entities.py backfill to clean up.

### DIFF
--- a/brain/src/hippo_brain/enrichment.py
+++ b/brain/src/hippo_brain/enrichment.py
@@ -3,7 +3,7 @@ import re
 import time
 import uuid
 
-from hippo_brain.entity_resolver import canonicalize, strip_worktree_prefix
+from hippo_brain.entity_resolver import canonicalize, is_path_type, strip_worktree_prefix
 from hippo_brain.models import EnrichmentResult, validate_enrichment_data
 from hippo_brain.watchdog import DEFAULT_LOCK_TIMEOUT_MS
 
@@ -97,14 +97,19 @@ def upsert_entities(conn, node_id: int, entities_dict, entity_type_map: dict, no
     entity_ids = []
     for key, entity_type in entity_type_map.items():
         for name in all_entities.get(key, []):
-            # Strip worktree prefix from the display name too — not just the
-            # canonical key. `canonical` deduplicates correctly, but `name` is
-            # what `mcp__hippo__get_entities` and the UI surface, and an
-            # ephemeral `.claude/worktrees/<X>/` prefix here means the first
-            # write from inside a worktree poisons the display name forever.
+            # For path-type entities, strip worktree prefix from the display
+            # name too — not just the canonical key. `canonical` deduplicates
+            # correctly, but `name` is what `mcp__hippo__get_entities` and the
+            # UI surface, and an ephemeral `.claude/worktrees/<X>/` prefix
+            # here means the first write from inside a worktree poisons the
+            # display name forever. Non-path entity types (errors stored as
+            # `concept`, etc.) are left verbatim — their values may legitimately
+            # contain `.claude/worktrees/...` substrings inside stack traces or
+            # diagnostic messages, and rewriting those would lose information
+            # while creating a name/canonical divergence.
             # On conflict, repair an existing polluted name with the new clean
             # one; otherwise leave it alone (avoid churn for stable rows).
-            display_name = strip_worktree_prefix(name)
+            display_name = strip_worktree_prefix(name) if is_path_type(entity_type) else name
             canonical = canonicalize(entity_type, name)
             cursor = conn.execute(
                 """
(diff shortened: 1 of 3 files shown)


stevencarpenter:
Addressed review feedback in 57e8725.

**1. Worktree-strip restricted to path types** (Codex r3150818224, Copilot r3150822944).

Both bots correctly flagged that `strip_worktree_prefix(name)` was running for *every* entity type, while `canonicalize` only does the same for path-types. Non-path entities — particularly errors stored as `concept` — can legitimately contain `.claude/worktrees/...` substrings inside diagnostic messages, and rewriting those would lose information while creating a name/canonical divergence.

Added a public `is_path_type(entity_type)` predicate in `entity_resolver.py` wrapping the existing `_PATH_TYPES` frozenset, and gated the strip on it:

```python
display_name = strip_worktree_prefix(name) if is_path_type(entity_type) else name
```

Non-path entities now pass through verbatim, matching what `canonicalize` does. New regression test `test_upsert_entities_does_not_strip_non_path_types` confirms an error message containing `.claude/worktrees/agent-XX/foo.py` survives into the `concept` entity row exactly as the LLM emitted it.

**2. Tests made deterministic** (Copilot r3150822955, r3150822964).

Both `test_upsert_entities_repairs_existing_polluted_name` and `test_upsert_entities_does_not_churn_clean_existing_name` assumed `canonicalize('file', '/Users/test/projects/hippo/brain/foo.py')` would reduce to `brain/foo.py`, which depends on `HIPPO_PROJECT_ROOTS` or auto-detection of `~/projects/hippo/.git`. Neither is guaranteed in CI.

Both tests now do:
```python
monkeypatch.setenv("HIPPO_PROJECT_ROOTS", "/Users/test/projects/hippo")
from hippo_brain.entity_resolver import _cached_fallback_roots
_cached_fallback_roots.cache_clear()
```

The cache-clear is in case a prior test in the same process exercised the auto-detect fallback.

**Test status:** 759 passed (up from 757 — the 6 new/touched tests are all deterministic), ruff check + format clean.

stevencarpenter:
### Code review

PR #106, opened 2026-04-28
fix(brain): dedup-entities also rebrands lone rows + two-pass UNIQUE safety

stevencarpenter:
Follow-up to #105. The first run of `dedup-entities.py` after #105 merged collapsed 626 duplicate rows but **left 146 worktree-polluted display names behind** — they were lone rows (no clean duplicate ever observed), and the script only operated on groups of size > 1. This PR closes that gap and also fixes a UNIQUE-constraint failure mode discovered while running it on the live DB.

### 1. Lone rows weren't processed at all

Before: the script grouped rows by `(type, new_canonical)` and merged any group with > 1 member. Single-member groups were silently skipped, even when the row's stored name/canonical was stale (worktree-polluted, missing project-root strip, missing `~`-expansion, etc.).

After: lone groups whose recomputed values differ from stored get UPDATEd in place. The new `_recompute_name()` helper mirrors the enrichment write path's display-name rule (strip worktree for path types, leave non-path types verbatim).

### 2. UNIQUE collision when one row's new canonical equals another row's current canonical

Discovered when the first attempt at running the patched script errored:

```
sqlite3.IntegrityError: UNIQUE constraint failed: entities.type, entities.canonical
```

This isn't a collision between two **new** canonicals — the grouping pass guarantees those are unique. It's the case where row A's new canonical equals row B's **current** canonical, and row B's *own* new canonical is something else (so they're in different groups). A naïve sequential UPDATE on row A trips the constraint even though the final state is collision-free.

Fix: two-phase placeholder pass.
- **Phase 1:** park every touched row at a unique placeholder (`__dedup_pending_<id>__`)
- **Phase 2:** apply the real new values

By Phase 2, no other row holds any of the target canonicals — UNIQUE always holds. Whole job stays in a single transaction with rollback.

## Live-DB result

| | Before this PR | After |
|---|---|---|
| Worktree-polluted display names | 146 | **31** |
| Total entities | 9,175 | 9,177 (+2 from concurrent brain writes) |
| Orphaned `knowledge_node_entities` | 0 | 0 |
| Placeholder leaks | n/a | 0 |
| Dupe groups | 0 | 0 |

The 31 remaining are correctly excluded:
- 20 `concept` entities — diagnostic error messages where the worktree path is information, not pollution (per #105 review)
- 10 `project` entities — not in `_PATH_TYPES`; LLM-extraction edge case worth a separate look
- 1 `file` entity literally named `.claude/worktrees/` — degenerate, regex correctly doesn't match

## Tests

The script previously had **zero** tests. This PR adds 6:

- `test_lone_polluted_row_gets_rebranded` — fresh fix; baseline new-feature coverage
- `test_lone_already_clean_row_not_touched` — no-op when nothing to do (avoid churn on every run)
- `test_non_path_lone_row_with_worktree_substring_preserved` — concept/error rows aren't lossy-stripped
- `test_duplicate_group_merge_also_strips_survivor_name` — merge survivor's `name` (not just canonical) is now re-canonicalized
- `test_rebrand_handles_canonical_swap_without_unique_collision` — **regression for the live failure**; would fail without the two-phase placeholder
- `test_dry_run_makes_no_changes` — `--dry-run` semantics

Loaded via `importlib.util` because the script's filename uses a hyphen.

## Test plan

- [x] `uv run --project brain pytest brain/tests` — 770 passed (was 764, +6 new)
- [x] Live-DB run: 382 lone rows rebranded, no errors, transaction committed cleanly
- [x] Verified post-run: 0 placeholder leaks, 0 orphaned links, brain healthy

## Notes / risks

- The two-phase placeholder pass adds N additional UPDATE statements where N = total rows touched. For 382 rows on the live DB, full job ran in ~4ms. Negligible.
- Placeholder values are namespaced + id-suffixed, so two parallel invocations can't collide on placeholders either (unlikely deployment scenario, but cheap defense).
- Survivor name update during merges is a behavior change — the previous version left the survivor's `name` polluted even after a successful merge. This was caught by the new `test_duplicate_group_merge_also_strips_survivor_name`.

### DIFF
--- a/brain/scripts/dedup-entities.py
+++ b/brain/scripts/dedup-entities.py
@@ -47,6 +47,7 @@ import logging
 import os
 import sqlite3
 import sys
+import uuid
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
@@ -187,15 +188,18 @@ def run(conn: sqlite3.Connection, dry_run: bool) -> dict[str, int]:
             final_updates.append((row["id"], new_name, new_canonical))
             total_rebranded += 1
 
-        # Phase 1: park every touched row at a unique placeholder canonical so
-        # the final UPDATEs can never collide on UNIQUE (type, canonical).
-        # Placeholder values are namespaced + id-suffixed, so two parallel
-        # invocations of this script wouldn't collide either (unlikely, but
-        # cheap to defend against).
+        # Phase 1: park every touched row at a placeholder canonical so the
+        # final UPDATEs can never collide on UNIQUE (type, canonical). The
+        # placeholder embeds a per-run UUID4 (32 hex chars) plus the entity id;
+        # collision with any real-world entity canonical or another in-flight
+        # invocation's placeholder is astronomically unlikely. Earlier versions
+        # used a deterministic `__dedup_pending_<id>__` token, which a
+        # malicious or coincidental entity row could collide with directly.
+        run_token = uuid.uuid4().hex
         for entity_id, _new_name, _new_canonical in final_updates:
             conn.execute(
                 "UPDATE entities SET canonical = ? WHERE id = ?",
-                [f"__dedup_pending_{entity_id}__", entity_id],
+                [f"__dedup_pending_{run_token}_{entity_id}__", entity_id],
             )
 
         # Phase 2: apply the real new values. Now no other entity row holds
--- a/brain/scripts/dedup-entities.py
+++ b/brain/scripts/dedup-entities.py
@@ -2,12 +2,21 @@
 """One-shot entity deduplication migration.
 
 Re-canonicalizes all entities using the current entity_resolver rules and
-merges rows where (type, new_canonical) would collide. Keeps the oldest row
-(smallest created_at), re-points all knowledge_node_entities to the survivor,
-then deletes the duplicate entity rows.
+either merges duplicates or rebrands lone rows whose stored values are stale.
 
-This script handles two flavors of duplication transparently because it
-delegates to canonicalize():
+For each (type, new_canonical) group:
+
+  * len > 1 — merge. Keep the oldest row (smallest created_at), re-point all
+    knowledge_node_entities to the survivor, delete the duplicate rows, and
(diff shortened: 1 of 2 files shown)


stevencarpenter:
Addressed all three review items in 7a2f794.

**1. Codex P2 — placeholder-collision risk** ([discussion r3150969342](https://github.com/stevencarpenter/hippo/pull/106#discussion_r3150969342))

Real concern. The deterministic `__dedup_pending_{id}__` token could legitimately collide with a pre-existing entity's canonical of the same type, tripping UNIQUE during the parking pass.

Fixed by embedding a per-run UUID4 (`__dedup_pending_{uuid_hex}_{id}__`). Collision with any real-world canonical is now astronomically unlikely. New regression test `test_placeholder_does_not_collide_with_existing_entity_canonical` sets up exactly the adversarial state Codex described — a `file` row whose canonical is `__dedup_pending_1__` plus a separate `file` row that needs rebranding (id=1) — and confirms the run completes cleanly under the new code. (The same setup would trip `sqlite3.IntegrityError` under the deterministic placeholder.)

**2. Copilot — `time.sleep(0.001)` flakiness** ([discussion r3150970292](https://github.com/stevencarpenter/hippo/pull/106#discussion_r3150970292))

Real concern — `_seed_entity` rounds to integer ms, so two calls within the same ms tick yield equal timestamps and merge-survivor selection becomes nondeterministic.

Fixed by adding an explicit `created_at` keyword to `_seed_entity` and passing deterministic timestamps 1 ms apart in the affected merge test. Docstring spells out when explicit timestamps are required. The `time.sleep` call is gone.

**3. Copilot — misleading dry-run comment** ([discussion r3150970303](https://github.com/stevencarpenter/hippo/pull/106#discussion_r3150970303))

Comment said "Stats reflect the work that *would* be done" but `run()` intentionally returns `rebranded=0`/`deleted=0` in dry mode. Adopted the suggested wording.

**Test status:** 766 passed (was 764, +1 new collision-regression test). Lint clean.

PR #107, opened 2026-04-28
fix(brain): project entities are path-typed; RAG synthesis preserves identifiers verbatim

stevencarpenter:
Two final #98 follow-up fixes that surfaced while verifying #104, #105, and #106 against the live DB. **No issues being deferred — same-session cleanup pass.**

## Fix 1: `project` entities are path-typed

The LLM consistently emits the work-session `cwd` as the project entity value, so values like `/Users/foo/projects/hippo` or `/Users/foo/projects/hippo/.claude/worktrees/agent-X` populate the `project` row's `name`. Because `project` was not in `_PATH_TYPES`, `canonicalize()` left those values un-stripped — every parallel-agent session yielded a distinct project entity polluted with `.claude/worktrees/<X>/` that never deduplicated with the clean root.

**Live-DB evidence** after PR #105 + #106's dedup pass:

```
SELECT type, COUNT(*) FROM entities WHERE name LIKE '%.claude/worktrees/%' GROUP BY type;
  concept | 20  ← intentionally preserved (diagnostic text per #105 review)
  file    |  1  ← degenerate `.claude/worktrees/` literal (regex correctly skips)
  project | 10  ← THIS FIX
```

Adding `project` to `_PATH_TYPES`:
- Strips worktree segments from project values.
- For values that match a project root exactly, returns the basename (`/users/foo/projects/hippo` → `hippo`).
- For non-matching paths (e.g. kafka-s3 when only hippo is in `HIPPO_PROJECT_ROOTS`), keeps the absolute form — still dedup-friendly across agents within the same project.

The existing `test_non_path_type_unaffected_by_path_logic` was using `project` as the example of "non-path type" — now points at `tool` (which is genuinely non-path), so the principle still holds for the truly-non-path types.

## Fix 2: RAG synthesis prompt preserves identifiers verbatim

PR #100 added a verbatim-preservation rule to the three enrichment prompts. The synthesis prompt in `rag.py` was never updated, so even when the new `qwen3.6-35b-a3b-ud-mlx` enrichment correctly preserved `HIPPO_FORCE` or `0.0.26` in `embed_text`, the synthesis layer paraphrased them away when answering RAG queries.

**Live-DB demonstration:** ran `mcp__hippo__ask` probe about `HIPPO_PROJECT_ROOTS`. The synthesis returned exact integer counts verbatim (`626 duplicates ... 146 lone rows ... 423 dupe groups`) — the new enrichment model preserved them — but missed the env-var name itself, returning *"I cannot identify the specific environment variable required..."* The literal token `HIPPO_PROJECT_ROOTS` was in the retrieved `embed_text`. Synthesis just didn't pull it forward because the prompt didn't say to.

The new rule in `_SYSTEM_PROMPT` mirrors the enrichment-prompt language verbatim: identifiers, version strings, env-var names, package@version pairs, file paths, CLI flags, exact integer counts, error codes — all preserved exactly. Hallucinated identifiers explicitly called out as worse than admitting "value not in context", with the `HIPPO_FORCE_INSTALL` vs `HIPPO_FORCE` example from #98.

## Tests

- `test_project_type_treated_as_path` — `/users/foo/projects/hippo` and `/users/foo/projects/hippo/.claude/worktrees/agent-X` both reduce to `hippo` when hippo is in `project_roots`.
- `test_project_type_strips_worktree_even_without_root_match` — worktree-strip applies even when `project_roots=[]` (kafka-s3 case).
- `test_system_prompt_contains_verbatim_preservation_rule` — synthesis prompt contains VERBATIM, version strings, environment variable, and hallucination warning.

## Test plan

- [x] `uv run --project brain pytest brain/tests` — **762 passed**, 1 skipped, 1 xfailed, 1 xpassed (was 759, +3 new tests)
- [ ] **Post-merge:** after this PR + #106 both land on main, re-run `brain/scripts/dedup-entities.py` to collapse the 10 worktree-polluted project entities into the clean `hippo`/`pp-bot` rows.
- [ ] **Post-merge:** re-run the F2 probe (`mcp__hippo__ask` about `HIPPO_PROJECT_ROOTS`) to verify synthesis now surfaces the env-var name verbatim.

## Expected post-merge state

```
SELECT type, COUNT(*) FROM entities WHERE name LIKE '%.claude/worktrees/%' GROUP BY type;
  concept | 20  ← intentionally preserved
  file    |  1  ← degenerate
  project |  0  ← cleaned
```

That brings the worktree-pollution metric to its true floor (21 entries, all intentionally preserved or degenerate).

## Out of scope (separate issue worth tracking)

Beyond worktree pollution, the `project` entity table has a *separate* mess of deeply-nested LLM-extraction artifacts (`hippo-eval/brain`, `hippo-roadmap/docs/capture-reliability`). Those are sub-paths the LLM wrongly tagged as project entities. Adding `project` to `_PATH_TYPES` doesn't fully fix them (they'd canonicalize to suffixes like `brain` if hippo-eval were a project root, or stay as absolute paths otherwise). Cleaner solution would be either (a) configuring those projects in `HIPPO_PROJECT_ROOTS`, or (b) tightening the enrichment prompt to forbid sub-path project values. Worth tracking but not blocking this cleanup pass.


stevencarpenter:
Addressed Codex P1 in 8d866c4.

**Codex P1 — cross-project canonical merge** ([discussion r3151011986](https://github.com/stevencarpenter/hippo/pull/107#discussion_r3151011986))

Real and important concern. Adding `project` to `_PATH_TYPES` made canonicalize apply both branches of the project-root logic — exact-match (good for projects) AND prefix-strip (unsafe for projects). So `/users/me/projects/hippo/brain` and `/users/me/projects/hippo-eval/brain` would both reduce to `brain` when both roots are configured, merging unrelated projects into one entity row.

Fixed by scoping the prefix-strip branch to non-project path types only. Exact-root match still applies to projects (`/users/me/projects/hippo` → `hippo`), so the worktree-cleanup case from this PR still works. Sub-paths (which the LLM sometimes emits as project values) stay verbatim — uglier but at least each project keeps a distinct identity.

The exact-match check is also reordered before the prefix-strip in the loop, so it fires on values like `/users/me/projects/hippo` (no trailing slash) before any prefix-strip would be attempted — verified with the existing `test_exact_root_match_returns_basename` still passing.

Two new tests:
- `test_project_type_does_not_merge_cross_project_subpaths` — direct regression for the failure mode. `hippo/brain` and `hippo-eval/brain` resolve to distinct canonicals.
- `test_file_subpath_dedup_still_works` — sanity check that the carve-out doesn't regress file-type behavior. `hippo/src/foo.rs` and `hippo-postgres/src/foo.rs` still merge to `src/foo.rs`.

**Test status:** 764 passed (was 762, +2 new tests). Lint clean.

PR #111, opened 2026-04-28
feat(brain): re-enrich existing knowledge nodes against current model + prompts

### LINKED ISSUE #110: Re-enrich existing knowledge nodes against current model + prompts (corpus uniformity)
## Background

Across PRs #100, #104, #105, #106, #107 (shipped in v0.19.1) we substantially upgraded what the enrichment pipeline produces:

- New `design_decisions` field on `EnrichmentResult` (PR #100).
- Verbatim-preservation rule in shell/claude/codex enrichment system prompts (PR #100).
- Identifier-dense `embed_text` (PR #100).
- Worktree-stripped entity display names (PR #105).
- Worktree-stripped + project-root-aware canonical for project entities (PR #107).
- Verbatim-preservation rule in RAG synthesis prompt (PR #107).

We also swapped the configured enrichment model from `gpt-oss-120b` to `qwen3.6-35b-a3b-ud-mlx` after dogfooding showed `gpt-oss-120b` paraphrased verbatim facts despite the prompt and pinned the GPU through a JSON-parse retry loop.

**The catch:** all of these only apply going forward. The existing 6,304 knowledge nodes (the entire corpus — verified via `SELECT COUNT(*) FROM knowledge_nodes`; oldest is 2026-04-01) were enriched against the *old* prompts and (for all but ~15) the *old* model. Specifically:

| Cohort | Approx count | What's missing |
|---|---|---|
| Pre-PR-100 nodes | ~4,500 | No `design_decisions` slot at all (defaults to empty array on read; field never populated) |
| Post-PR-100 / gpt-oss-120b | ~1,790 | `design_decisions` slot present but model paraphrased verbatim facts; embed_text quality below new prompt's standard |
| qwen3.6 (post-swap) | ~15 | Properly verbatim — these are the only "good" rows |

Result: **>99% of recall queries hit content that paraphrased away the verbatim facts.** No synthesis-time prompt fix can recover what wasn't preserved at enrichment time. The source events (in `events`, `claude_session_segments`, `browser_events`) still have the raw text — re-enriching against the new prompts + model would meaningfully improve recall accuracy across the whole corpus.

## Scope

All 6,304 knowledge nodes. The corpus is < 30 days old, so there's no "ancient stuff not worth re-doing" tier — it's all recent and all queryable.

## Core design question: replace vs. supersede vs. update-in-place

Three options. Recommendation up front: **update in place.**

### Update in place (recommended)

Keep the same `knowledge_nodes.id`, re-derive `content` / `summary` / `embed_text` / `entities` / `design_decisions` / `tags` / `outcome` from the source events, atomically apply via UPDATE inside a transaction. Re-compute and re-upsert the embedding.

Pros:
- No orphaned `knowledge_node_entities`, `knowledge_node_events`, `knowledge_node_claude_sessions`, `knowledge_node_browser_events`, `knowledge_node_workflow_runs`, or `knowledge_node_lessons` rows — they all reference the existing id.
- `created_at` preserved (still reflects when the work *happened*).
- `uuid` preserved (any cached references stay valid).
- FTS5 / sqlite-vec triggers fire on UPDATE and re-index correctly.

Cons:
- Embedding regeneration is the dominant cost.
- Need to bump `enrichment_version` (existing schema column, currently always 1) so resume-after-interrupt logic can skip already-processed nodes.
- Slight risk if the new enrichment errors mid-flight — the old content is gone. Mitigated by per-node transaction.

### Replace (delete + re-insert)

Same as update-in-place semantically but generates a new `id` + `uuid`.

Cons:
- Cascades to every link table — must explicitly re-link to the new id.
- Loses `created_at` unless we copy it forward.
- Harder to make resumable.

Wins nothing over update-in-place that I can see.

### Supersede (keep both, mark old)

Add a `superseded_by_id INTEGER` column. Keep old nodes for history; queries filter `WHERE superseded_by_id IS NULL`.

Pros:
- Auditable rollback path if the new enrichment somehow regresses.
- "What did the old answer look like" is preserved.

Cons:
- Doubles knowledge_nodes storage (and thus FTS5 + vec0 index size).
- Schema migration.
- Every retrieval query needs the filter (easy to forget).
- The "history" is weak justification: nodes are *derived* from events; events are the source of truth and are unchanged. Re-derivation is idempotent.

I'd skip supersede.

## Implementation sketch

New CLI subcommand: `hippo brain re-enrich [--all | --since=<date> | --where=<sql>] [--dry-run] [--limit=N]`

Or a one-shot script: `brain/scripts/re-enrich-knowledge-nodes.py` (parallels `dedup-entities.py`).

Core logic:
1. Read knowledge_nodes WHERE `enrichment_version < CURRENT_VERSION` (use a constant in code; bump when prompts change non-trivially).
2. For each, look up the source events via the link tables (`knowledge_node_events`, `knowledge_node_claude_sessions`, etc.).
3. Re-build the prompt using the same `build_*_enrichment_prompt` functions the live brain uses.
4. Call `_call_llm_with_retries` against the configured enrichment model.
5. UPDATE the knowledge node with the new content / embed_text / entities. Use the same `upsert_entities` path so worktree-strip + project canonicalization apply.
6. Re-embed and UPSERT into `knowledge_vectors`.
7. Bump `enrichment_version` for resume-after-interrupt.

Run order: newest-first by `created_at` (so active queries benefit earliest) but configurable. Stagger or yield occasionally so real-time enrichment of new events doesn't starve.

Throughput estimate: brain currently sustains ~tens of nodes/min in steady state on `qwen3.6-35b-a3b-ud-mlx`. 6,304 nodes ≈ a few hours of background work. Run overnight.

## Acceptance criteria

- [ ] All knowledge nodes have `enrichment_version` ≥ the new constant.
- [ ] `SELECT COUNT(*) FROM knowledge_nodes WHERE NOT json_valid(content)` stays at the workflow-summary baseline (~168) — re-enrichment doesn't introduce new malformed JSON.
- [ ] Random sample of pre-#100 nodes now have populated `design_decisions` arrays where the events warrant it.
- [ ] `hippo ask "what dependency bumps shipped in v0.13.0?"` returns `python-multipart 0.0.22 → 0.0.26` and `pygments 2.19.2 → 2.20.0` verbatim. (This is the canonical F2 probe from #98 — currently fails against the live DB because the relevant nodes were enriched by gpt-oss-120b.)
- [ ] No orphaned `knowledge_node_*` link rows after the run.
- [ ] FTS5 + sqlite-vec indexes consistent with `knowledge_nodes` (no rowid mismatches).
- [ ] Idempotent: running twice in a row is a no-op on the second run.

## Risks

- **Cost.** Several GPU-hours. Acceptable as a one-time + occasional-after-prompt-changes operation. Not acceptable as a per-edit hook.
- **Bad new enrichment.** If `qwen3.6-35b-a3b-ud-mlx` is somehow worse on a specific class of events than `gpt-oss-120b` was, re-enrichment makes things worse. Mitigation: spot-check a small batch first (`--limit=50 --dry-run` then `--limit=50` for real, then probe).
- **Live brain interference.** Re-enrichment runs in addition to the steady-state enrichment loop. They'd contend for the same LM Studio. Mitigation: run with explicit nice/throttle, or temporarily stop the brain (`mise run stop`), re-enrich offline, then `mise run start`.
- **Workflow nodes.** `workflow_enrichment.py` writes free-text markdown into `content` (not JSON). Re-enrichment script must route those through `_enrich_workflow_*` rather than the JSON-EnrichmentResult path. Worth handling explicitly to avoid corrupting them.

## Out of scope

- Changing the prompts further. Today's prompts are the target; re-enrichment is what bridges the corpus to them.
- Supersede semantics (see above).
- Re-enrichment as a hot path / continuous backfill. Single-shot operation.

## Related

- Closes the operational tail of #98 — the prompt + model + DB-state changes shipped, but the existing corpus stayed paraphrased.
- Independent of #108 (RAG truncation) — even with #108 fixed, the underlying enrichment quality is what re-enrichment lifts.

stevencarpenter:
Closes #110.

One-shot script that re-runs enrichment for nodes that pre-date the prompt + model upgrades shipped in v0.19.1. **Update-in-place** semantics — preserves `id` / `uuid` / `created_at` / link rows, bumps `enrichment_version` from 1 → 2 so re-runs resume cleanly.

## Sources handled

- **shell** (~2,447 nodes) — events looked up via `knowledge_node_events`, prompt via `build_enrichment_prompt`.
- **claude** (~3,895 nodes) — segments via `knowledge_node_claude_sessions`; prompt = `\n---\n`-joined `claude_sessions.summary_text`, mirroring `Server._enrich_claude_batches`.

Together ~99% of the corpus. **Browser (6) and workflow (187) stay at v1** — workflow needs a separate path because its `content` is free-text markdown, not a JSON `EnrichmentResult`; browser is too small to justify the third branch in this script.

## CLI

```
re-enrich-knowledge-nodes.py [--db PATH] [--source shell|claude|all]
                             [--limit N] [--dry-run] [--throttle-ms MS]
                             [--newest-first | --oldest-first]
                             [--model MODEL]
```

Default behavior: process newest first (active queries benefit early), 200ms throttle between nodes (so the live brain's real-time enrichment isn't starved), all sources.

## Design choices (from #110)

- **Reuses brain functions.** Prompt builders, `parse_enrichment_response`, `upsert_entities`, `embed_knowledge_node` are imported directly. The script is a wrapper, not a re-implementation.
- **UPDATE in place.** Inside one transaction: UPDATE `knowledge_nodes`, DELETE `knowledge_node_entities` for this node, re-call `upsert_entities` (which routes through current `canonicalize` + `strip_worktree_prefix`).
- **Bump version only on success.** Failures stay at v1, retried on next run.
- **Resume.** Default WHERE filters `enrichment_version < TARGET`; running twice in a row is a no-op on the second run.

## Tests (5)

- `test_candidate_selection_respects_version_filter` — v2 nodes excluded.
- `test_candidate_selection_source_filter` — `--source` flag isolates.
- `test_process_node_updates_in_place` — uuid/created_at preserved, version bumps, content replaced.
- `test_dry_run_makes_no_changes` — no LLM call, no DB mutation.
- `test_failed_node_keeps_old_version_for_retry` — failure path keeps v1.

Hyphenated filename → loaded via `importlib.util` (same pattern as `test_dedup_entities.py`).

## Test plan

- [x] `uv run --project brain pytest brain/tests` — 776 passed, 1 skipped, 1 xfailed, 1 xpassed
- [x] Live-DB dry-run with `--limit 5` — selected mix of claude/shell candidates with reasonable prompt sizes; no errors.
- [ ] Live-DB real run with `--limit 5` to spot-check end-to-end output quality before kicking off the full corpus pass.
- [ ] Operator runs full pass overnight after merge.

## Operator playbook

```bash
# Verify the model + LM Studio are happy
hippo doctor

# Spot-check on a small batch first
HIPPO_PROJECT_ROOTS=\$HOME/projects/hippo \\
  uv run --project brain python brain/scripts/re-enrich-knowledge-nodes.py \\
    --limit 5

# Inspect the result
sqlite3 ~/.local/share/hippo/hippo.db "
  SELECT id, length(embed_text), substr(embed_text, 1, 200)
  FROM knowledge_nodes WHERE enrichment_version = 2
  ORDER BY updated_at DESC LIMIT 5;
"

# If output looks right, run the full pass
HIPPO_PROJECT_ROOTS=\$HOME/projects/hippo \\
  uv run --project brain python brain/scripts/re-enrich-knowledge-nodes.py
```

Throughput on `qwen3.6-35b-a3b-ud-mlx`: ~tens of nodes/min. Full corpus (~6,300 nodes) ≈ a few hours background. Throttle defaults to 200ms; bump higher if real-time enrichment appears starved.

## Acceptance criteria (from #110)

- [x] All processed nodes have `enrichment_version` ≥ TARGET (= 2).
- [x] No orphaned `knowledge_node_*` link rows after the run (the script never deletes anything but `knowledge_node_entities`, which is re-populated via `upsert_entities`).
- [x] Idempotent: second run is a no-op (verified via the version-filter test).
- [ ] Random sample of pre-existing nodes have populated `design_decisions` after re-enrichment (operator verification).
- [ ] `hippo ask` returns verbatim version strings for v0.13.0 dep-bump probe (operator verification).

## Risks (from #110)

- **Cost.** Several GPU-hours. One-shot operation; not a per-edit hook.
- **Bad new enrichment.** If a specific event class produces worse output under qwen3.6 than under the original model, re-enrichment makes things worse. Mitigation: spot-check `--limit 5` before kicking off full corpus.
- **Live brain interference.** Re-enrichment + steady-state enrichment contend for the same LM Studio. Mitigation: 200ms throttle; bump if needed; or stop the brain (`mise run stop`) for fully-offline run.

### DIFF
--- a/brain/scripts/re-enrich-knowledge-nodes.py
+++ b/brain/scripts/re-enrich-knowledge-nodes.py
@@ -1,0 +1,449 @@
+#!/usr/bin/env python3
+"""One-shot re-enrichment of existing knowledge nodes.
+
+Re-runs enrichment for knowledge nodes that pre-date the current prompt /
+model setup, using the same code paths the live brain uses, and updates
+each node in place. Preserves ``id`` / ``uuid`` / ``created_at`` / link
+rows; only the derived content (summary, embed_text, entities, tags,
+design_decisions, etc.) is replaced.
+
+Why this exists
+---------------
+Across PRs #100, #105, #107 the enrichment system prompts were upgraded
+(verbatim preservation, identifier-dense embed_text, design_decisions
+schema, worktree-stripped entity names). The configured enrichment model
+also moved from ``gpt-oss-120b`` to ``qwen3.6-35b-a3b-ud-mlx``. None of
+that retroactively benefits the existing corpus. Running this script
+brings every previously-enriched node up to the current standard.
+
+Sources handled
+---------------
+- **shell** — link table ``knowledge_node_events``, prompt via
+  ``build_enrichment_prompt`` against rows from ``events``.
+- **claude** — link table ``knowledge_node_claude_sessions``, prompt is
+  the concatenation of ``claude_sessions.summary_text`` rows joined by
+  ``\\n---\\n``, mirroring ``Server._enrich_claude_batches``.
+
+Sources NOT handled by this version
+-----------------------------------
+- **workflow** — ``workflow_enrichment.py`` writes free-text markdown
+  into ``knowledge_nodes.content`` rather than a structured JSON
+  EnrichmentResult; needs a separate code path. 187 nodes, low priority.
+- **browser** — only 6 nodes in the live corpus; not worth a third
+  branch in this script.
+
+Both stay at ``enrichment_version=1`` and will be skipped by future runs
+of this script. Add explicit handling later if needed.
+
+Usage
+-----
+    uv run --project brain python brain/scripts/re-enrich-knowledge-nodes.py \\
+        [--db PATH] [--source shell|claude|all] [--limit N] [--dry-run] \\
+        [--throttle-ms MS] [--newest-first]
+
+Resume semantics
+----------------
+The script bumps each successfully-processed node's
+``enrichment_version`` from 1 to 2. The default WHERE clause filters
+``enrichment_version < TARGET_VERSION``, so re-running picks up where a
+prior run left off. Failed nodes stay at the old version and will be
+retried on the next run.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import logging
+import os
+import sqlite3
+import sys
+import time
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT / "src"))
+
+from hippo_brain.claude_sessions import CLAUDE_SYSTEM_PROMPT  # noqa: E402
+from hippo_brain.client import LMStudioClient  # noqa: E402
+from hippo_brain.embeddings import embed_knowledge_node  # noqa: E402
+from hippo_brain.enrichment import (  # noqa: E402
+    SHELL_ENTITY_TYPE_MAP,
+    SYSTEM_PROMPT,
+    build_enrichment_prompt,
+    parse_enrichment_response,
+    upsert_entities,
+)
+
+# CLAUDE entity-type map mirrors SHELL — claude enrichment writes the same
+# structured EnrichmentResult shape into knowledge_nodes.content.
+CLAUDE_ENTITY_TYPE_MAP = SHELL_ENTITY_TYPE_MAP
+
+TARGET_ENRICHMENT_VERSION = 2
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s %(levelname)s %(name)s: %(message)s",
+    stream=sys.stderr,
+)
+log = logging.getLogger("re-enrich")
+
+
+def _default_db_path() -> Path:
+    base = os.environ.get("XDG_DATA_HOME") or str(Path.home() / ".local" / "share")
+    return Path(base) / "hippo" / "hippo.db"
+
+
+def _load_settings() -> dict:
+    """Pull lmstudio + model settings from the canonical brain config."""
+    from hippo_brain import _load_runtime_settings
+
+    return _load_runtime_settings()
+
+
+def _select_candidate_nodes(
+    conn: sqlite3.Connection, source: str, limit: int | None, newest_first: bool
+) -> list[dict]:
+    """Return knowledge_nodes rows that need re-enrichment, joined with source.
+
+    ``source`` is one of ``shell`` / ``claude`` / ``all``. The returned dicts
+    carry an explicit ``_source`` key naming the link table the row hit.
+    """
+    order_clause = "DESC" if newest_first else "ASC"
+    limit_clause = f"LIMIT {int(limit)}" if limit and limit > 0 else ""
+
+    queries = []
+    if source in ("shell", "all"):
+        queries.append(
+            f"""
+            SELECT n.id, n.uuid, n.created_at, 'shell' AS _source
+            FROM knowledge_nodes n
+            JOIN knowledge_node_events kne ON kne.knowledge_node_id = n.id
+            WHERE n.enrichment_version < {TARGET_ENRICHMENT_VERSION}
+            GROUP BY n.id
+            """
+        )
+    if source in ("claude", "all"):
+        queries.append(
+            f"""
+            SELECT n.id, n.uuid, n.created_at, 'claude' AS _source
+            FROM knowledge_nodes n
+            JOIN knowledge_node_claude_sessions kncs ON kncs.knowledge_node_id = n.id
+            WHERE n.enrichment_version < {TARGET_ENRICHMENT_VERSION}
+            GROUP BY n.id
+            """
+        )
+
+    union_sql = " UNION ".join(queries)
+    final_sql = f"SELECT * FROM ({union_sql}) ORDER BY created_at {order_clause} {limit_clause}"
+    rows = conn.execute(final_sql).fetchall()
+    return [dict(r) for r in rows]
+
+
+def _fetch_shell_events(conn: sqlite3.Connection, node_id: int) -> list[dict]:
+    """Return event rows linked to the given knowledge node, ordered by timestamp."""
+    rows = conn.execute(
+        """
+        SELECT e.id, e.session_id, e.timestamp, e.command, e.exit_code, e.duration_ms,
+               e.cwd, e.git_branch, e.git_commit, e.git_repo, e.stdout, e.stderr,
+               e.hostname, e.shell
+        FROM events e
+        JOIN knowledge_node_events kne ON kne.event_id = e.id
+        WHERE kne.knowledge_node_id = ?
+        ORDER BY e.timestamp ASC
+        """,
+        (node_id,),
+    ).fetchall()
+    return [dict(r) for r in rows]
+
+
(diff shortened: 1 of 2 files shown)


stevencarpenter:
Addressed the actionable review items in 984c490.

## Items now resolved

**1. JSON-hint on retry** ([Copilot r3151469189](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151469189))

`_call_llm_with_retries` now appends *"Your previous response was not valid JSON. Output ONLY a JSON object, no explanation or markdown."* on attempts ≥ 2, exactly mirroring `Server._call_llm_with_retries` at [server.py:909-918](https://github.com/stevencarpenter/hippo/blob/main/brain/src/hippo_brain/server.py#L909-L918). Without this hint, a model that emits prose around the JSON would burn 3 full inferences in a row.

**2. Unused `CLAUDE_ENTITY_TYPE_MAP`** ([Copilot r3151469207](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151469207))

Removed the alias rather than threading a per-source map through. Claude and shell share the entity-type map by design (both write `EnrichmentResult` into `knowledge_nodes.content`), so a registry would silently no-op today and the cleanest fix is to delete the unused symbol.

**3. Missing claude round-trip test** ([Copilot r3151469226](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151469226))

Added `test_process_claude_node_updates_in_place` — seeds a `claude_sessions` row, processes via `_source='claude'`, asserts the node was updated AND the joined `summary_text` reached the LLM (so the claude prompt branch is actually exercised, not silently skipped).

## Items already fixed before review (false-positive on stale snapshot)

The two P1 findings ([Codex r3151457304](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151457304) / [Copilot r3151469160](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151469160) on sqlite-vec loading; [Codex r3151457307](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151457307) / [Copilot r3151469179](https://github.com/stevencarpenter/hippo/pull/111#discussion_r3151469179) on version-bump-before-embed) were addressed in [dea6cc6](https://github.com/stevencarpenter/hippo/pull/111/commits/dea6cc6) before the bot reviews landed.

Both fixes are visible in the current head:
- `vector_store.open_conn(db_path)` is used to load sqlite-vec ([line ~290 of the script](https://github.com/stevencarpenter/hippo/blob/984c490/brain/scripts/re-enrich-knowledge-nodes.py#L290)).
- `_finalize_node_version` is a separate function called only after both `_update_node_in_place` AND `_re_embed` succeed; `_update_node_in_place` itself does NOT touch `enrichment_version`. The docstring explicitly notes the rollback semantics.

So both P1 concerns are correctly defended against in the current head, and the existing `test_failed_node_keeps_old_version_for_retry` test pins the resume-after-failure behavior.

**Tests:** 6 passed (was 5, +1 claude round-trip). Lint + format clean.

PR #112, opened 2026-04-28
feat: RAG entity surfacing + env_var entity type (closes #108) — v0.20.0

### LINKED ISSUE #108: RAG synthesis misses identifiers present in retrieved embed_text (per-hit truncation in _hit_lines)
## Background

PR #107 added a verbatim-preservation rule to the RAG synthesis prompt (`brain/src/hippo_brain/rag.py::_SYSTEM_PROMPT`). Combined with the model swap to `qwen3.6-35b-a3b-ud-mlx`, this fixed two of the original Issue #98 F2 failure modes:

- **Numeric facts preserved exactly.** `gpt-oss-120b` used to paraphrase `python-multipart 0.0.22 → 0.0.26` to *"patched two transitive Python vulnerabilities"*. Today's probe returned `626 deletions / 382 rebrands / 558 to 31` verbatim.
- **"Admit unknown" replaces "hallucinate plausible variant."** The original failure was hallucinating `HIPPO_FORCE_INSTALL` when the real value was `HIPPO_FORCE`. Today's probe explicitly says *"context does not mention any specific environment variable... please share additional logs"* instead of guessing.

But there's a remaining gap that's NOT a prompt-side problem.

## The bug

Synthesis claims missing info for identifiers that are demonstrably present in the retrieved `embed_text`.

**Probe asked:** *"What specific environment variable does brain/scripts/dedup-entities.py require, and what is the placeholder canonical format used in Phase 1?"*

**Synthesis answered:** *"the context does not mention any specific environment variable required by brain/scripts/dedup-entities.py... nor does it describe a 'Phase 1' update flow or specify the placeholder canonical format used."*

**But the retrieved nodes contain both literally.** Querying the live DB:

```sql
SELECT substr(embed_text, 1, 250) FROM knowledge_nodes
WHERE enrichment_model = 'qwen3.6-35b-a3b-ud-mlx'
ORDER BY created_at DESC LIMIT 5;
```

returns rows like:

```
6295 | hippo brain/scripts/dedup-entities.py brain/tests/test_dedup_entities.py
       brain/tests/conftest.py fix/dedup-entities-lone-rows main git checkout main
       git pull --ff-only origin main uv run --project brain ruff check ruff format
       pytest HIPPO_PROJECT_ROO...
```

`HIPPO_PROJECT_ROOTS` is right there in `embed_text`. Synthesis didn't surface it.

## Root cause hypothesis

`brain/src/hippo_brain/rag.py::_hit_lines`:

```python
if hit.get("embed_text"):
    lines.append(f"Detail: {_truncate(hit['embed_text'], embed_cap)}")
```

The per-hit `embed_cap` is allocated proportionally by `_allocate_payload_caps` against a `DEFAULT_MAX_CONTEXT_CHARS = 8000` budget split across however many hits + their `commands_raw` and `design_decisions`. With 10 retrieval hits and the budget split across 3 fields per hit, individual `embed_text` payloads can get truncated to a few hundred characters or less.

`embed_text` is laid out as a tag-soup of identifiers (PR #100 F4 made it intentionally identifier-dense), but the order is non-prioritized — symbols at the head get surfaced, symbols at the tail get truncated away. `HIPPO_PROJECT_ROO[TS]` is literally cut off mid-token in the sample above.

So: prompt rule says "preserve identifiers verbatim," retrieval delivers identifiers in `embed_text`, but truncation drops them before they reach the synthesis model. Synthesis is following the rule correctly given what it sees.

## Repro

1. Find a recent qwen-enriched knowledge node whose `embed_text` contains a specific identifier:
   ```sql
   SELECT id, embed_text FROM knowledge_nodes
   WHERE enrichment_model = 'qwen3.6-35b-a3b-ud-mlx'
     AND embed_text LIKE '%HIPPO_PROJECT_ROOTS%'
   LIMIT 1;
   ```
2. Run `hippo ask "what env var does dedup-entities.py require?"` (or via MCP).
3. Observe: synthesis claims the env var isn't in context.
4. Inspect the retrieved hits' actual `embed_text` length vs. the truncated rendering passed to the LLM (instrument `_build_rag_prompt` to log the final user-message char count and the per-hit `embed_text` post-truncation).

### Option A — bigger embed_text caps

Bump `DEFAULT_MAX_CONTEXT_CHARS` and/or change the proportional allocation in `_allocate_payload_caps` to favor `embed_text` over `commands_raw`. Risk: blows past the synthesis model's effective context window when many large hits return.

### Option B — identifier-priority truncation

Replace the simple `_truncate(text, max_len)` for `embed_text` with a smarter strategy: extract identifier tokens (UPPER_CASE, version regex, `pkg@ver`, file paths, function names) and rank them ahead of generic words. Truncate the tail when budget runs out, but never drop an identifier from the head. Risk: regex-based identifier extraction is fragile; needs careful design and tests.

### Option C — surface structured entities in the rag prompt

`knowledge_nodes` has structured `entities.tools`, `entities.files`, etc. (per PR #100). These are already canonicalized identifiers. Render them as a separate `Identifiers: <comma-list>` line in `_hit_lines` so the synthesis model sees them outside the prose-truncation budget. Risk: increases context size; could push duplicates with what's already in `embed_text`.

I'd start with **C** as the lowest-risk path. The identifiers are already structured and small; rendering them at the top of each hit's context block (before the truncatable Detail line) gives synthesis a guaranteed surface to bind to.

## Acceptance criteria

- [ ] An automated test demonstrates the failure (a hit whose `embed_text` contains an identifier-string at a position past the truncation cap; synthesis must reproduce it in the answer).
- [ ] After the fix, `hippo ask "what env var does dedup-entities.py require?"` returns `HIPPO_PROJECT_ROOTS` verbatim against the current live DB.
- [ ] The fix doesn't regress existing rag tests.

## Out of scope

- Changing the synthesis model or its prompt — those changed in PR #107 and the *prompt-following* behavior is correct now.
- Changing what the enrichment model writes into `embed_text` — PR #100 F4 already made that identifier-dense.

## Related

- Closes follow-up gap from #98 F2 (synthesis-side).
- Logical successor to #107 — that PR fixed the prompt; this fixes the retrieval pipeline.

stevencarpenter:
## Summary

Closes #108. Two waves of work bundled into one PR:

**Wave 1 — RAG entity surfacing (structural fix).** Identifier-typed entities (`tool`, `file`, `service`, `project`, `env_var`) hydrated on retrieval and rendered as a dedicated `Entities:` line above the truncatable `Detail:` block in synthesis prompts. The old `embed_text` proportional truncation was clipping identifiers mid-token; the new line lives in the structural budget and is preserved verbatim through truncation.

**Wave 2 — `env_var` as a first-class entity type.** The pre-impl probe (`IMPL_NOTES.md` Gate 0) showed `HIPPO_PROJECT_ROOTS` was bucketed only as a `concept` entity, so wave 1 alone wouldn't close the canonical repro. Wave 2 adds `env_var` to the entities CHECK list (schema v12→v13), updates all three enrichment prompts (shell, claude_sessions, browser) to extract `env_vars` arrays, and bumps `CURRENT_ENRICHMENT_VERSION` to 3 to invalidate the corpus for re-enrichment.

## Schema migration: v12 → v13

Extends `entities.type` CHECK list with `'env_var'`. SQLite cannot ALTER a CHECK constraint in place, so the migration follows the documented 12-step recipe:

1. `PRAGMA foreign_keys = OFF` (outside transaction)
2. `BEGIN`
3. `DROP TABLE IF EXISTS entities_new` (idempotency for partial-success crashes)
4. `CREATE TABLE entities_new` with the extended CHECK
5. Copy all rows
6. `DROP TABLE entities`
7. `ALTER TABLE entities_new RENAME TO entities`
8. Recreate `idx_entities_type_name` and `idx_entities_canonical`
9. `PRAGMA user_version = 13` (inside the same `execute_batch` for atomicity)
10. `COMMIT`
11. `PRAGMA foreign_keys = ON`

Tested against happy-path migration of all existing entity types + crash-recovery (half-populated `entities_new` from a prior crashed attempt).

## Breaking change: coordinated daemon-first upgrade

`ACCEPTED_READ_VERSIONS` collapsed from `{12, 10, 9, 8, 7, 6, 5}` to `{13}`. Brain now writes `env_var`-typed entity rows on every node, and v5–v12 schemas all carry the pre-`env_var` CHECK. Connecting brain to any older DB would crash the enrichment loop with `CHECK constraint failed: entities`. Reject at connect time instead.

**Upgrade order:** daemon first (runs the v12→v13 migration on startup), then brain. The single-host invariant makes this safe in practice; the daemon-side `schema_handshake.rs` already enforces strict equality.

## Operational notes

- **Active re-enrichment was killed.** A run was at 232/6343 (~3.7%) when this work began. Per-node-transaction model means no half-state. After this PR lands, restart re-enrichment from scratch — the `enrichment_version < 3` filter will select every node, including the 232 that completed pre-bump (they still need the env_var bucket).
- **`CURRENT_ENRICHMENT_VERSION = 3`** is the new ratchet. Now imported as `TARGET_ENRICHMENT_VERSION` in `re-enrich-knowledge-nodes.py` so a future bump invalidates the corpus automatically.
- **`hippo ask "what env var does dedup-entities.py require?"`** is the canonical repro — should return `HIPPO_PROJECT_ROOTS` verbatim once re-enrichment completes.

## Single source of truth for entity-type taxonomy

Two new tuples in `enrichment.py`:

- `IDENTIFIER_ENTITY_TYPES = ("tool", "file", "service", "project", "env_var")` — surfaced on the RAG `Entities:` line.
- `NON_IDENTIFIER_ENTITY_TYPES = ("concept", "domain")` — deliberately excluded.

`brain/tests/test_entity_taxonomy.py` discovers every top-level `*_ENTITY_TYPE_MAP` constant via `pkgutil.walk_packages` and asserts every value is classified. Adding a new entity type without updating one of the tuples now fails this test loudly. `BROWSER_ENTITY_TYPE_MAP` was extracted from an inline literal so the guard covers it too.

## Truncation note (review feedback)

The `Entities:` line uses **token-boundary truncation**, not char-based. Char-based `_truncate` would emit `HIPPO_PROJECT_RO…`, which is exactly the mid-identifier failure mode this PR exists to prevent. The packer reserves space for a `", …"` ellipsis tail and drops whole tokens from the tail when over budget. Pathological single-token-exceeds-cap case falls back to char truncation on that one token so the line still surfaces something rather than vanishing.

## Files updated for v0.20.0

| File | 0.19.1 → 0.20.0 |
|---|---|
| `Cargo.toml` (`workspace.package.version`) | manual |
| `brain/pyproject.toml` | manual |
| `hippo-gui/VERSION` | manual |
| `Cargo.lock` | regenerated |
| `brain/uv.lock` | regenerated |

Minor (not patch) because: new entity type, schema bump, breaking `ACCEPTED_READ_VERSIONS` change.

## Test plan

- [x] `cargo test --workspace` — all green (includes new v12→v13 migration tests + crash-recovery test)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `uv run --project brain pytest brain/tests` — 795 passed, 1 skipped, 1 xfailed, 1 xpassed
- [x] `uv run --project brain ruff check brain/src brain/tests` — clean
- [x] `uv run --project brain ruff format --check brain/src brain/tests` — clean
- [ ] After merge: rebuild + reinstall daemon, restart brain, kick off re-enrichment from scratch, verify `hippo ask "what env var does dedup-entities.py require?"` returns `HIPPO_PROJECT_ROOTS` verbatim


stevencarpenter:
## Review feedback addressed in ff63d10

Three real findings from the bot reviews — all addressed:

**1. Codex P2 — `_fetch_details` non-deterministic entity order** (`brain/src/hippo_brain/retrieval.py:560`)
> The inner `ROW_NUMBER() OVER (... ORDER BY ent.type, ent.name)` only orders rows within each partition; the outer query had no `ORDER BY`. SQLite is free to return rows in any plan-dependent sequence.

Fixed by adding explicit `ORDER BY knowledge_node_id, type, name` on the outer query so the rendered `Entities:` line is stable across runs.

**2. Copilot — `test_entity_taxonomy.py` import side effects** (`brain/tests/test_entity_taxonomy.py:48`)
> Imports every module under `hippo_brain` to discover `*_ENTITY_TYPE_MAP` constants — triggers `server.py`'s `logging.basicConfig(...)` at import time and other module-level init.

Replaced `pkgutil.walk_packages` with an explicit `_MODULES_WITH_ENTITY_MAPS` allowlist: only `enrichment` + `browser_enrichment` are imported. The contract is now that a future map module extends the tuple in the same diff.

**3. Copilot — `validate_enrichment_data` silently dropped `domains` bucket** (`brain/src/hippo_brain/models.py:79`)
> `_ENTITY_KEYS` doesn't include `domains`, so any model-emitted `entities.domains` is silently dropped before `upsert_entities` sees it.

This was a **pre-existing bug**, not introduced by this PR. Verified on main: `SELECT type, COUNT(*) FROM entities GROUP BY type` shows zero `domain`-typed rows despite the browser source having run for weeks. Fixed by adding `"domains"` to `_ENTITY_KEYS` and the `EnrichmentResult.entities` default factory. Added regression test `test_parse_enrichment_response_preserves_domains_bucket` that would have caught this on day one of the browser source.

Tests: 796 passed (+1 regression), full Rust workspace green, ruff + clippy clean.

stevencarpenter:
## Bot re-review pass on ff63d10 — all three findings are stale

The post-push bot pass re-emitted the same three comments. Confirmed against the actual `ff63d10` tree — all three are already addressed:

**Codex P2 on `brain/src/hippo_brain/retrieval.py:568` ("no outer ORDER BY")**
The outer `ORDER BY` is on line 560, two lines above where Codex pointed:
\`\`\`
559        WHERE rn <= 20
560        ORDER BY knowledge_node_id, type, name
561        \"\"\"
\`\`\`
Hydration order is now deterministic (alphabetical by type then name within each node).

**Copilot on `brain/tests/test_entity_taxonomy.py` ("imports every module via pkgutil.walk_packages")**
This comment is pinned to commit `57a405e` (the older commit). `ff63d10` replaced the walker with an explicit `_MODULES_WITH_ENTITY_MAPS` allowlist — see lines 40–53. No more whole-package import side effects.

**Copilot on `brain/src/hippo_brain/models.py:80` ("`_ENTITY_KEYS` does not include `domains`")**
Already fixed at line 79:
\`\`\`python
_ENTITY_KEYS = ("projects", "tools", "files", "services", "errors", "env_vars", "domains")
\`\`\`
Plus `"domains": []` added to `EnrichmentResult.entities` default factory, plus a regression test (`test_parse_enrichment_response_preserves_domains_bucket`).

All three were addressed in `ff63d10`; the post-push re-review appears to have run against a stale diff baseline rather than the current tree state.

stevencarpenter:
## Subagent code review findings addressed in 1089522

Ran `superpowers:code-reviewer` on the full PR diff (`8083b10..ff63d10`). Verdict was **APPROVE** with five minor findings — all addressed:

| Finding | Fix |
|---|---|
| **M1** — `_render_entities_line` lost the omission signal in the (cap - tail, cap] borderline case | Fallback now appends a bare `…` for multi-token cases, signals omission without mid-token cuts. New test `test_entities_borderline_first_token_still_signals_omission`. |
| **M2** — `_to_result` defensive branch (when `detail is None`) didn't pass `entities=` explicitly | Added for stylistic consistency with the rest of the branch. Behavior was already correct via the dataclass default. |
| **M3** — v13 migration missing SQLite's recommended `PRAGMA foreign_key_check` between RENAME and COMMIT | Added. Redundant for today's FK shape (textual name resolution post-rename) but defensive against future schema evolution. |
| **M4** — no test for the partial-schema "no entities table" else branch | New `test_migrate_v12_to_v13_advances_version_when_entities_missing`. |
| **M5** — `# 4 chars: ", …"` comment was wrong (the string is 3 chars; math used `len()` so behavior was right) | Replaced with a note that lengths come from live `len()` calls. |

Tests: 797 brain (+1), 6 hippo-core migration tests (+1), full Rust workspace + ruff + clippy clean.

(Also re-checked the PR's bot review threads — same three Codex/Copilot comments from earlier are still pinned to old commits; all addressed in `ff63d10`. No new bot activity.)

PR #141, opened 2026-05-07
fix(brain): surface LM Studio response body on 4xx/5xx

stevencarpenter:
## Summary

- LM Studio returns a JSON body on 4xx (e.g. `{"error": "Context history must not be empty."}`) that pinpoints the failure. The previous code discarded it via httpx's default `raise_for_status`, leaving `brain.stderr.log` with only `400 Bad Request for url ...`.
- Add `_raise_with_body()` helper that appends up to 500 chars of the response body to the `HTTPStatusError`, applied to `chat()`, `embed()`, and `list_models()`.
- Keeps the existing exception type so callers (`_call_llm_with_retries`, queue-failure handlers) need no changes.

## Why this matters

While auditing capture health, we found 21 × `400 Bad Request` errors against `/v1/chat/completions` in the last brain session — 10 on `workflow.enrich`, 1 on `shell.chat` (after retries), 1 hidden under `_call_llm_with_retries` retry warnings on the claude path. The same `run_id` 400s once and 200s on a subsequent attempt within seconds, so we know it's transient.

We don't yet know what LM Studio is actually saying because the body was thrown away. This PR is the prerequisite for a properly informed root-cause investigation — once it's deployed, future 400s will include LM Studio's reason in the brain log, and we can decide whether the next fix is (a) bring workflow's chat call under the same retry wrapper as the other paths, (b) tune model preflight, or (c) something the body reveals that we haven't considered.

## Test plan

- [x] `uv run --project brain pytest brain/tests/test_client.py brain/tests/test_client_real.py -v` — 18 passed (3 new)
- [x] `uv run --project brain pytest brain/tests` — 883 passed, 1 skipped, 1 xfailed, 1 xpassed
- [x] `uv run --project brain ruff check brain/src brain/tests` — clean
- [x] `uv run --project brain ruff format --check brain/src brain/tests` — clean
- [ ] Deploy and observe brain.stderr.log next time a 4xx fires — confirm body appears in the log

## New tests added

- `test_chat_400_includes_response_body_in_error` — asserts the LM Studio error reason now appears in `HTTPStatusError`
- `test_embed_400_includes_response_body_in_error` — same for embeddings
- `test_chat_error_with_empty_body_does_not_blow_up` — empty-body 503 still raises cleanly without a synthetic `Body:` suffix

### DIFF
--- a/brain/src/hippo_brain/client.py
+++ b/brain/src/hippo_brain/client.py
@@ -32,10 +32,15 @@ def _raise_with_body(resp: httpx.Response) -> None:
     # LM Studio returns a JSON body on 4xx (e.g. {"error": "Context history must
     # not be empty."}) that pinpoints the failure. httpx's default raise_for_status
     # discards it, so we re-raise with the body appended to keep diagnoses visible.
+    # If body extraction itself fails (decode error, body unread, etc.), fall back
+    # to the original raise — never let a body-extraction error mask the real HTTP error.
     try:
         resp.raise_for_status()
     except httpx.HTTPStatusError as e:
-        body = resp.text[:500].strip()
+        try:
+            body = resp.text[:500].strip()
+        except Exception:
+            body = ""
         if not body:
             raise
         raise httpx.HTTPStatusError(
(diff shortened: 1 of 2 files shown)


stevencarpenter:
### Code review

PR #148, opened 2026-05-10
feat(migration): update references from LM Studio to inference server (oMLX)

stevencarpenter:


stevencarpenter:
### Code review

Found 1 issue:

1. The class rename `LMStudioClient` → `InferenceClient` (no compatibility alias) breaks three top-level operational scripts in `scripts/` that still import the old name. Each script will raise `ImportError` at the module-import stage on first execution. The PR updated `brain/scripts/` but missed `scripts/`.

Rename site: https://github.com/stevencarpenter/hippo/blob/9158fa61866209b23019dcb1b90a5c62c51a62d4/brain/src/hippo_brain/client.py#L64-L66

Stale imports:

- https://github.com/stevencarpenter/hippo/blob/9158fa61866209b23019dcb1b90a5c62c51a62d4/scripts/hippo-bulk-enrich.py#L23-L25
- https://github.com/stevencarpenter/hippo/blob/9158fa61866209b23019dcb1b90a5c62c51a62d4/scripts/hippo-re-embed.py#L21-L23
- https://github.com/stevencarpenter/hippo/blob/9158fa61866209b23019dcb1b90a5c62c51a62d4/scripts/hippo-ingest-claude.py#L144-L146

Fix: either update those three call sites to `InferenceClient`, or add `LMStudioClient = InferenceClient` as a deprecation alias in `client.py` next to the legacy `[lmstudio]` config shim.

---

A handful of lower-confidence findings did not clear the review-gate threshold but are worth a glance:

- `docs/brain-watchdog.md` still extensively documents `hippo.brain.lmstudio.crashes` and the "model has crashed" body-match logic that this PR removed from `client.py` (the `_lm_crashes` counter from #143). The doc and the code are now inconsistent.
- `otel/grafana/dashboards/hippo-enrichment.json` panel title `"LM Studio Errors / min"` was not renamed to match the new `hippo_brain_inference_errors_total` query.
- `brain/src/hippo_brain/server.py` line 268 comment says `inference_reachable` "may be added in a future release", but the very next line emits it.
- `brain/src/hippo_brain/bench/` was not migrated — `cli.py`, `coordinator.py`, `preflight.py` still default to `:1234` / use `lmstudio_*` field names. Bench is acknowledged as already-incomplete, so this may be intentionally deferred.

stevencarpenter:
Closing this PR — the vendor-neutrality sweep has been absorbed into PR #149 in commit [ceb1fb5](https://github.com/stevencarpenter/hippo/commit/ceb1fb541b18cf66c92c8eb53e3a4f9d22050a1c).

Why fold into #149 rather than the other way around: #149 was already in a tested known-good state with brain healthy locally, schema migrations live (v14), and the I-11/I-12 watchdog invariants + health-grade dashboard panel landed. Rebasing #149 onto a freshly-merged #148 would have replayed every conflict twice. Folding the smaller sweep into the larger feature PR is one round of conflict resolution instead of two.

What was preserved from #148:
- All 24 files exclusive to this PR were taken verbatim (AGENTS.md, README.md, CLAUDE.md, brain/README.md, the website pages, mise.toml, scripts/*, brain/tests/test_evaluation*.py, brain/tests/test_rag.py, brain/tests/test_workflow_enrichment.py, brain/tests/conftest.py, brain/src/hippo_brain/rag.py, brain/src/hippo_brain/bench/enrich_call.py, brain/scripts/migrate-v5-to-v6.py, etc.).
- #148's deeper rename on shared files was adopted where it went further than #149's pass (evaluation.py, mcp.py field renames, workflow_enrichment.py kwarg, server.py prose polish).

Where the two PRs diverged on policy, #149's policy wins (each was discussed with the user when the divergence surfaced):
- **Fail-loud on legacy `[lmstudio]` config section.** #148 silently accepted it as fallback; #149 raises RuntimeError with a migration message. The original incident (silent fallback to localhost:1234 when the user's config said `[inference]` at port 8000) is exactly what fail-loud prevents.
- **Default inference URL stays `http://localhost:1234/v1`** to match `config/config.default.toml`. The shipped template default doesn't change without a user-visible flag day.
- **`/health` exposes `inference_reachable` only.** The daemon's `hippo doctor` parse falls back to the legacy `lmstudio_reachable` for cross-version safety, so old brains running against a new daemon still report correctly.

Plus one cross-doc drift fix #148 inherited from main and #149 fixed: CLAUDE.md said "I-1..I-10" — bumped to "I-1..I-12" so it matches `docs/capture/architecture.md`.

Thanks for doing the surface sweep — the 24 docs/website/test files would have been easy to miss in the feature PR otherwise.

PR #86, opened 2026-04-25
feat(watcher): T-5 — Claude session FS watcher + schema v10 (P2.1)

stevencarpenter:
This pull request introduces schema version 10, adding support for a new Claude session ingestion mode using a file system watcher, along with related configuration, migration, and testing updates. It also adds new dependencies and commands to facilitate this functionality. The most important changes are summarized below.

**Database Schema and Migration:**
[...]

PR #149, opened 2026-05-11
feat(opencode): add opencode session ingestion via SQLite polling

stevencarpenter:
Introduces schema v14 with `agentic_sessions`, `agentic_enrichment_queue`, and `agentic_cursor` tables to support opencode (and future harness) live ingestion alongside Claude sessions, plus a new `hippo opencode-poll` command and `[opencode]` config block.


stevencarpenter:
### Code review

Found 7 issues:

1. `open_opencode_db` opens the opencode DB with `SQLITE_OPEN_READ_ONLY` and then calls `pragma_update(None, "journal_mode", "WAL")`. Setting journal mode requires writing the DB header; on a read-only connection SQLite returns `SQLITE_READONLY` and the `?` propagates, so every `poll_tick` fails before any data is read. Drop the WAL pragma on the read-only connection.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L181-L194

2. The polling cursor is advanced with `latest.time_updated` but the read query (and the cursor field name) filter on `time_created`. Any session whose `time_created` falls between the prior `time_created` and `time_updated` is permanently skipped — silent data loss whenever `time_updated > time_created` (the normal case). Should be `last_time_created: latest.time_created`.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L255-L262

3. `write_opencode_knowledge_node` constructs `INSERT INTO knowledge_node_agentic_sessions (knowledge_node_id, agentic_session_id) VALUES (?, ?, ...)` with one placeholder per segment id and then binds `(node_id, *segment_ids)` — `N` placeholders against `N+1` params. The junction table has exactly two columns, so the only correct shape is a multi-row insert with `(node_id, seg_id)` pairs (or a loop). As written this raises a `ProgrammingError` on every call with at least one segment.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/opencode_sessions.py#L310-L319

4. `poll_tick` hardcodes the Hippo DB path as \`dirs::home_dir().join(\".local/share/hippo/hippo.db\")\`, ignoring \`XDG_DATA_HOME\`. CLAUDE.md says: \"All paths use XDG defaults... Override with \`XDG_DATA_HOME\` / \`XDG_CONFIG_HOME\` env vars.\" The rest of the daemon resolves this through \`config.db_path()\` / \`hippo_core::config\`. Resolve from config instead.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L216-L223

5. The cursor is upserted unconditionally after the upsert loop, even when every `upsert_session` errored. If a batch fails entirely (schema mismatch, lock, etc.), \`errors_sent\` increments but the cursor still moves past the rows — the next poll's \`WHERE time_created > ?1\` excludes them forever. Skip the cursor advance (or abort the tick) when \`errors_sent > 0\`.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L238-L262

6. The brain's enrichment dispatch in \`server.py\` gathers shell, claude, browser, and workflow — but nothing imports or calls \`opencode_sessions\`, and no code reads/writes \`agentic_enrichment_queue\`. Rows the Rust poller writes into \`agentic_sessions\` sit at \`enriched = 0\` indefinitely. Opencode enrichment is shipped but never runs.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/server.py#L938-L946

7. The new \`agentic_sessions\` table has no \`probe_tag\` column, unlike \`events\`, \`claude_sessions\`, and \`browser_events\` (all of which carry one for AP-6 compliance). \`docs/capture/adding-a-source.md\` step 2 says to \"match the existing patterns: ... \`probe_tag TEXT\` for filtering synthetic canaries,\" and \`docs/capture/anti-patterns.md\` AP-6 forbids probe rows in event tables \"without filtering them from every user-facing query.\" Add \`probe_tag TEXT\` to the DDL and gate the \`source_health\` update on \`probe_tag IS NULL\`.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-core/src/schema.sql#L549-L575

stevencarpenter:
### Code review (addendum)

Five additional lower-confidence findings (each scored ≥40 but below my normal ≥80 cutoff — treat as secondary):

1. `upsert_session` writes the \`agentic_sessions\` row and the \`source_health\` row in two separate \`conn.execute()\` calls with no transaction. The capture-reliability contract in [\`docs/capture/architecture.md\`](https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/docs/capture/architecture.md) requires \"each path writes to its source's events table AND to \`source_health\` in the same SQLite transaction.\" If the daemon crashes (or the DB locks) between the two statements, the row lands without the corresponding \`source_health\` update, breaking watchdog freshness. Wrap both in \`conn.unchecked_transaction()\` like \`claude_session.rs::insert_segments\` does.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L132-L178

2. \`insert_opencode_segment\` inserts into a table named \`opencode_segments\` that doesn't exist anywhere in the v14 schema (the actual table is \`agentic_sessions\`). The function is currently unreferenced — no caller anywhere in \`brain/\` — so it doesn't fire today, but it's a latent landmine. Delete the function (it looks like leftover scaffolding from an earlier table name).

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/opencode_sessions.py#L223-L264

3. The \`source_health\` write is a bare \`UPDATE … WHERE source = 'agentic-session-opencode'\`. If the row is missing (e.g., the migration's \`has_source_health\` guard skipped seeding on a test DB), the UPDATE silently affects 0 rows — no error, no health write, and the watchdog has nothing to assert on. Prefer \`INSERT … ON CONFLICT(source) DO UPDATE\` so the capture path is self-healing.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L167-L175

4. Fresh-install \`schema.sql\` seeds **two** rows (\`agentic-session-claude\` and \`agentic-session-opencode\`), but the v13→v14 migration in \`storage.rs\` only seeds \`agentic-session-opencode\`. Existing installs upgrading to v14 will permanently lack the \`agentic-session-claude\` row, while fresh installs will have it. Add the second \`INSERT OR IGNORE\` to the migration to keep fresh-install and upgrade-install invariants aligned.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-core/src/schema.sql#L613-L621

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-core/src/storage.rs#L720-L731

5. \`OpenConfig\` has an \`enabled\` field (default \`true\`), but \`Commands::OpencodePoll\` calls \`poll_tick\` unconditionally without checking it — the flag is dead. Mirror the existing watchdog pattern a few lines up that gates \`watchdog::run\` on \`config.watchdog.enabled\`.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/main.rs#L932-L972

stevencarpenter:
### Code review (final addendum, low-confidence)

Posting the remaining 4 candidates for completeness. Confidence here is much lower than the previous batches — treat as discussion points rather than action items:

1. \`iter_pending_opencode_segments\` marks rows as \`enriched = 1\` *before* enrichment runs (the actual LLM call happens later in the still-unwired pipeline). If the LLM call fails, the segment stays at \`enriched = 1\` and is never retried. The claude analogue uses a \`status\` machine (\`processing\` → \`done\`/\`failed\`) and the v14 schema even ships an \`agentic_enrichment_queue\` table for exactly this — but nothing uses it. Currently dormant because of finding #6 in my first comment (the pipeline is dead), but it would activate the moment opencode enrichment is wired up. Confidence: 25.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/opencode_sessions.py#L72-L120

2. \`make_source_key\` uses the inode on macOS but falls back to a constant \`\"opencode-0\"\` on \`#[cfg(not(target_os = \"macos\"))]\`. The inode keying exists so cursors stay distinct across DB replacement / reinstall; the Linux fallback silently collapses all DBs onto one cursor key. Since hippo is macOS-only in production this only affects Linux CI tests (where it can mask cursor contamination between runs). Confidence: 25.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L75-L89

3. The v13→v14 migration uses \`if version == EXPECTED_VERSION - 1 || (version > 0 && version < EXPECTED_VERSION && version != EXPECTED_VERSION - 1)\` — which simplifies to \`(1..=13).contains(&version)\`, the idiom every prior migration uses. There's also a duplicated comment block immediately above it (likely a merge artifact), and the \`has_source_health\` probe is hoisted outside the guard so it runs on every \`open_db\` call regardless of version. Pure style / maintainability — no functional bug, just deviates from the canonical pattern. Unscored, dropped during consolidation as a nitpick.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-core/src/storage.rs#L647-L663

4. \`poll_tick\` has no \`tokio::time::timeout\` (or any other deadline) wrapping the opencode-DB open / cursor read / per-row upsert loop. If the opencode DB hangs (lock, network FS, etc.), the launchd oneshot blocks indefinitely. The launchd \`KeepAlive\`/throttle eventually triggers a restart, so this is a soft issue — but the watcher-side equivalent (PR #86) flagged the same gap. Unscored, dropped as speculative.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L196-L270

Note: I said \"extra 5\" in the previous request but only 4 candidates remained after the addendum batch; not inventing a fifth.

stevencarpenter:
### Code review — second-opinion pass

After the prior review, I dispatched a fresh reviewer with no knowledge of my findings to read the diff start-to-finish. It surfaced several real issues my targeted passes missed. Crediting the independent pass:

1. \`upsert_session\` populates \`summary_text\` with \`s.directory\` and \`source_file\` with \`s.id\`. \`summary_text\` is the enrichment-input column the brain prompt builder reads; with the cwd path in that column, the LLM prompt becomes \"Opencode session at /Users/foo/proj. Snapshot diffs: +0/-0 lines, 0 files\" — no actual content. Knowledge nodes would be uniformly useless even if the pipeline ran.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L155-L165

2. There is no launchd plist for opencode (no \`com.hippo.opencode-*.plist\`) and \`Commands::OpencodePoll\` is a one-shot tick — \`config.opencode.poll_interval_secs\` is dead config, read by nothing. The CLI doc says \"LaunchAgent, OneShot\" but no agent ships. \`mise run install\` won't start it; \`hippo doctor\` won't notice it isn't running. The feature does not function on a real install without an additional plist or cron.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L196-L270

3. \`make_source_key\` returns \`String::new()\` if \`metadata()\` fails. The empty string is then used as the primary key in \`agentic_cursor\`. A transient I/O error makes the daemon write to a \`source_key = \"\"\` row, and a subsequent successful poll writes to \`opencode-<inode>\` — splitting cursor state across two rows and re-replaying sessions.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L75-L89

4. \`is_enrichment_eligible(seg, \"opencode\")\` falls through to the default \"unknown source, default eligible\" branch — there is no opencode-specific eligibility filter. Every session (including trivial ones with 0 messages and 0 diffs) gets routed to the LLM. \`docs/capture/adding-a-source.md\` step 6 mandates a per-source branch.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/opencode_sessions.py#L168-L185

5. \`iter_pending_opencode_segments\` claims segments by flipping \`agentic_sessions.enriched = 1\` directly — the new \`agentic_enrichment_queue\` table the v14 migration ships is never written or read by any Python code. The claude path uses a proper claim → lock → retry through \`claude_enrichment_queue\` for exactly this reason. As written, any enrichment failure (LLM error, network blip) loses the segment forever.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/brain/src/hippo_brain/opencode_sessions.py#L72-L145

6. \`data_version\` is read into \`_data_version\` and never used. The surrounding comment claims \"skip if no new writes since last tick\" — but no comparison happens. Either implement the gate (store \`last_data_version\` per \`source_key\`, skip the \`session\`-table read when unchanged) or delete the dead read.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-daemon/src/opencode_session.rs#L210-L218

7. \`schema.sql\` line 621 contains shipped LLM debris: \`-- This topic is exposed at the /ask endpoint as well. (e.g.,). M i. .\` — leftover from an editing pass. This is production schema.

https://github.com/stevencarpenter/hippo/blob/1689ce76e4a9b9facfb2f083d5b81ceeee2fdd69/crates/hippo-core/src/schema.sql#L619-L624

The second-opinion reviewer's verdict: **Not ready to merge.** Concurring assessment.

stevencarpenter:
### Code review (delta `1689ce7..a837e2f`)

Found 3 issues in the fix-pass / deferred-items closeout. Note: I authored these commits, so this is a self-review surfaced by the parallel-reviewer subagents.

1. The new I-11 watchdog invariant is structurally dead — it fires on `consecutive_failures > 3`, but `opencode_session::upsert_session` never bumps `consecutive_failures` on error or resets it on success. The only code that touches that column is `daemon.rs::flush_events`, which the opencode poller doesn't use. Result: the column stays at `DEFAULT 0` forever and the proxy can never trigger.

https://github.com/stevencarpenter/hippo/blob/a837e2f3fb19d5809f8931142a95195ab6b6932b/crates/hippo-daemon/src/watchdog.rs#L578-L598

2. Related: the opencode `source_health` write omits four fields the `claude_session.rs` anchor sets — `consecutive_failures = 0`, `last_success_ts`, `events_last_1h`, `events_last_24h`. Without these, the doctor has no event-rate signal and the watchdog has no failure-counter to gate on (this is what makes #1 unreachable). The anchor pattern lives at \[claude_session.rs:1102-1119\](https://github.com/stevencarpenter/hippo/blob/a837e2f3fb19d5809f8931142a95195ab6b6932b/crates/hippo-daemon/src/claude_session.rs#L1102-L1119).

https://github.com/stevencarpenter/hippo/blob/a837e2f3fb19d5809f8931142a95195ab6b6932b/crates/hippo-daemon/src/opencode_session.rs#L237-L249

3. `adding-a-source.md` §8 says: \"Add an entry to the I-1..I-N invariant list in `crates/hippo-daemon/src/watchdog.rs` **and the matching documentation in `architecture.md`**.\" I-11 is in watchdog.rs but missing from `docs/capture/architecture.md`, and the watchdog.rs file-level doc comment still reads \"Asserts invariants I-1..I-10\" — stale once I-11 ships.

https://github.com/stevencarpenter/hippo/blob/a837e2f3fb19d5809f8931142a95195ab6b6932b/crates/hippo-daemon/src/watchdog.rs#L1-L13

stevencarpenter:
### Code review (delta `b5387b9..664f9b04`)

Found 1 issue:

1. Grafana panel grid overlap. The \`Embedding Failures\` panel (id 8) was widened from \`w: 3\` to \`w: 6\` at \`x: 6\`, so its right edge now extends to column 12. The adjacent \`RAG Latency by Stage\` panel (id 9) still sits at \`x: 9, w: 8\` — three columns (9–11) overlap. Grafana renders overlapping panels visually stacked, so the row at \`y: 16\` is broken. Fix: shift panel 9 to \`x: 12\` and shrink/shift its neighbors (panel 10 currently \`x: 17, w: 7\`) to fit within the 24-column grid.

https://github.com/stevencarpenter/hippo/blob/664f9b04373c9978de9470fa7ef5d589e205e836/otel/grafana/dashboards/hippo-enrichment.json#L439-L451

Lower-confidence findings that didn't clear the bar:

- \`nan_count\` in the instrumented \`_vec_blob\` is dead diagnostic info — \`struct.pack(\"<f\", float(\"nan\"))\` succeeds, so NaN never triggers the error path the helper instruments. Replace with \`inf_count\` (also packs fine but more useful upstream) or drop the field (scored 50).

- \`isinstance(x, (int, float))\` correctly skips packable elements (ints/bools coerce to float in \`struct.pack\`), so the int-in-tuple concern is academic (scored 5).

PR #176, opened 2026-05-25
fix(brain): stop infinite re-enrichment of un-enrichable claude segments

stevencarpenter:
## Summary

Fixes an infinite enrichment loop on a single Claude session segment, plus the broader bug class behind it.

A subagent transcript (`claude_sessions.id=4592`) made the local LLM emit invalid JSON (`Invalid \escape`). Enrichment failed parsing on every attempt. Because `last_enriched_content_hash` only advanced on **success**, the watcher's dedup gate (`content_hash != last_enriched_content_hash`) never closed, so the segment was re-enqueued forever — the `max_retries=5` cap was reset on each re-enqueue. Over ~18h this burned ~263 enrichment cycles / ~1300 inference calls on one segment.

## Changes

- **Terminal-failure watermark** (`mark_claude_queue_failed`): on a terminal failure (retries exhausted → `failed`), advance `last_enriched_content_hash = content_hash`, closing the dedup gate for un-enrichable content. Self-healing: if the content later changes, `content_hash` diverges and the segment gets a fresh attempt.
- **JSON-escape repair** (`parse_enrichment_response`): repair invalid backslash escapes and retry parsing once before raising. Most such responses now enrich instead of being abandoned. (Verified: segment 4592 now enriches to a valid node on a single claim.)
- **Skip-path gate closure** (`_skip_ineligible_claude_segments`): same class — `skipped` segments also never advanced the watermark, so they re-enqueued on every restart/file-change (churn + a duplicate node per legacy segment). Skipping now advances the watermark too (`COALESCE` preserves it for legacy NULL-hash rows).

## Test plan

- [x] `pytest brain/tests` — 897 passed, 1 skipped, 1 xfailed
- [x] `ruff check` + `ruff format --check` clean
- [x] New unit tests: terminal-failure watermark (+ NULL-hash safety), JSON-escape repair (+ valid-escape preservation), skip-path gate closure (+ NULL-hash safety)
- [x] Live verification after deploy: segment 4592 enriched to a valid node, `retry_count=0`, gate `CLOSED`; `JSONDecodeError=0` and zero claude enrichment-failed loops post-restart

## Out of scope (filed separately)

While verifying, found a pre-existing, unrelated bug: ~1152/1499 workflow (`change_outcome`) knowledge nodes have invalid-JSON content (raw model chain-of-thought stored instead of parsed JSON, since ~April). Not addressed here.

### DIFF
--- a/brain/src/hippo_brain/claude_sessions.py
+++ b/brain/src/hippo_brain/claude_sessions.py
@@ -622,8 +622,16 @@ def _skip_ineligible_claude_segments(conn, segments: list[dict]) -> list[dict]:
                 "WHERE claude_session_id = ?",
                 (reason, now_ms, seg["id"]),
             )
+            # Advance the dedup watermark so the watcher does not re-enqueue this
+            # segment on every restart/file-change. Skipping is terminal — the
+            # eligibility verdict is deterministic — so an open gate is pure churn
+            # (and a duplicate node per legacy segment on re-run). COALESCE keeps
+            # an existing watermark untouched when content_hash is NULL (legacy).
             conn.execute(
-                "UPDATE claude_sessions SET enriched = 1 WHERE id = ?",
+                "UPDATE claude_sessions "
+                "SET enriched = 1, "
+                "    last_enriched_content_hash = COALESCE(content_hash, last_enriched_content_hash) "
+                "WHERE id = ?",
                 (seg["id"],),
             )
     if len(eligible) != len(segments):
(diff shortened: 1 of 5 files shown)


stevencarpenter:
Thanks — all three addressed:

- **Watermark race (lines 793 & 635)**: confirmed against the daemon (`insert_segments` upserts `content_hash` unconditionally on reparse; `decide_enqueue` only suppresses re-enqueue while `processing`). Fixed both paths to advance `last_enriched_content_hash` only to the **attempted** hash, gated on `content_hash` still equalling it — so newer content the worker never saw can't be stranded. `mark_claude_queue_failed` now takes `content_hashes`; the skip path uses the claimed `seg['content_hash']`. Added race-regression tests for both.
- **`_VALID_JSON_ESCAPE` comment**: corrected to state it matches a valid escape OR a lone backslash.

PR #177, opened 2026-05-25
fix(brain): workflow enrichment writes validated JSON + records model

stevencarpenter:
## Summary

Fixes a pre-existing, ongoing data-quality bug: workflow (`change_outcome`) knowledge nodes stored raw LLM output instead of validated JSON.

`enrich_one_async` wrote the raw `inference.chat(...)` reply straight into `knowledge_nodes.content` with no parse/validate, and omitted `enrichment_model` from the INSERT. A reasoning model's chain-of-thought ("Here's a thinking process: ...") therefore landed verbatim in `content`. Live DB shows **~1152/1499 (77%)** of `change_outcome` nodes have invalid-JSON content (dating to ~April), which breaks `json_extract` in RAG/search.

## Changes

- New `WORKFLOW_SYSTEM_PROMPT` instructs a JSON object (mirrors the claude path's schema).
- `enrich_one_async` now sends `[system, user]` messages and routes the reply through `parse_enrichment_response` (validate + escape-repair), serializing the validated `EnrichmentResult` into `content`.
- On un-parseable output it raises → **no garbage node** is written and the run is marked failed (retryable) instead.
- `enrichment_model` is now recorded on the INSERT. `embed_text` (the `repo@sha` title) and the authoritative `outcome` column are unchanged.

## Test plan

- [x] `pytest brain/tests` — 893 passed, 1 skipped, 1 xfailed
- [x] New tests: stores valid JSON + records model; rejects non-JSON output (no node written)
- [x] Updated existing workflow/server tests whose mocks returned non-JSON
- [x] `ruff check` + `ruff format --check` clean

## Notes / follow-ups (not in this PR)

- **Backfill** of the existing ~1152 invalid nodes is intentionally separate (needs signoff): delete bad nodes (FTS auto-cleans via trigger; `knowledge_vectors` vec0 needs a manual `DELETE` first — no cascade), then reset `workflow_runs.enriched=0` + queue→`pending` to re-enrich.
- **Dead code**: the sync `enrich_one` calls `inference.complete`, which `InferenceClient` does not implement — it's broken/dead and only referenced by tests. Left for separate removal.
- Composes with #176 (escape-repair in `parse_enrichment_response`); no overlapping files.

### DIFF
--- a/brain/src/hippo_brain/workflow_enrichment.py
+++ b/brain/src/hippo_brain/workflow_enrichment.py
@@ -9,6 +9,7 @@ For each completed workflow run in the enrichment queue:
 6. For each failing annotation, call lessons.upsert_cluster
 """
 
+import json
 import logging
 import sqlite3
 import time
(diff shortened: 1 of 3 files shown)


stevencarpenter:
Thanks for the review. Addressing the three comments:

- **tags column not populated** and **embed_text discarded** — both already fixed in 83f0761 (pushed before this review landed): the INSERT now writes `tags`, `enrichment_version`, and `embed_text = result.embed_text` (mirroring the claude path), and the dead sync `enrich_one` was deleted.
- **"repairs stray escapes" comment** — corrected to say strips fences + validates JSON (the escape-repair lands with #176, not this branch).

PR #188, opened 2026-05-29
fix(enrichment): stop knowledge-node duplication + dedup the backlog

stevencarpenter:
## What & why

The enrichment pipeline minted a **fresh** `knowledge_nodes` row on every re-enrichment with no replacement. When the watcher re-enqueued a grown/re-parsed agentic segment, enrichment **appended** an identical node instead of replacing the prior one — so long-lived sessions accreted byte-identical duplicates. State at investigation time:

- **2,486 redundant nodes (~18% of the corpus)**, e.g. one `13cc2867` segment carried **14 byte-identical** nodes minted over 121 minutes.
- The class was **still active** (~110 new dups/day), so cleaning data alone would have let it regrow.

Completeness was separately audited and found **clean**: Cursor and Codex fully captured (Codex's one "missing" thread is a *deleted-on-disk* rollout — unrecoverable, not a hippo gap), and the only absent Claude session is a 0-turn empty stub (correctly absent). No reconcile action was needed — the remediation is purely dedup + stopping regrowth.

## The fix (defense in depth)

**Stop the class (agentic writers):**
- **opencode daemon content-hash gate** (`decide_enqueue`), matching codex/cursor — opencode previously had *no* gate. Brain advances `last_enriched_content_hash` for opencode.
- **`replace_prior_agentic_nodes`** — write-time replacement of the prior `observation` node(s) for a re-enriched segment (all `knowledge_node_*` link rows + the vec0 vector, in the same transaction), wired into the claude + opencode writers. **Scoped to `node_type='observation'`** so the workflow enricher's co-linked `change_outcome` (CI-outcome) nodes are never collateral damage.
- **skip-unchanged claim guard** — drop a claimed segment whose `content_hash == last_enriched_content_hash` before the LLM call.
- `_get_conn` now loads `sqlite_vec` so the write path can delete vectors atomically (and closes the conn if the load fails, to avoid leaking + double-faulting the error handlers).

**Remediate the backlog:**
- `brain/scripts/dedup-knowledge-nodes.py` — one-shot global dedup keyed on `(content, embed_text, node_type)`; `MIN(id)` survivor with a chaining guard; unions all 7 auto-discovered `knowledge_node_*` link tables onto the survivor; deletes loser vectors + rows; `foreign_key_check` gated before COMMIT; **dry-run default**.

**Guard against regression:**
- **Watchdog I-16** — alarms on any agentic segment carrying >1 `observation` node with identical `(content, embed_text, node_type)`. Scoped to `observation` (the class the gate guards).
- **`hippo doctor`** — new Claude + Cursor session-coverage checks; the Codex deleted-rollout case demoted from `[!!]` to informational.

## Verification

- **Tests:** `cargo fmt`/`clippy -D warnings`/`test -p hippo-daemon` (incl. 4 new I-16 tests) green; `cargo build --release --features otel` green. Brain: **936 passed**, ruff clean. New: `test_dedup_knowledge_nodes.py` (11 cases) + `test_agentic_node_replacement.py` (incl. the change_outcome-survival regression guard from review).
- **Adversarial review:** a multi-agent review caught a genuine BLOCKER (node_type-blind replacement would have silently deleted 168 CI-outcome nodes) and a docs over-claim — both fixed before merge.
- **Live remediation executed:** stopped all writers → checkpoint + APFS backup → dry-run → `--apply`. Result: **13,517 → 11,031 nodes (−2,486)**, 11,031 vectors, **0 orphan vectors, 0 unembedded, 0 orphan link rows across all 7 tables, FK-clean, I-16 = 0**, idempotent re-run = 0. Fix deployed; post-restart enrichment creates no new dups.

## Follow-up (tracked, not in this PR)

The **workflow + browser** enrichers share the append-only pattern and need their own replacement gate keyed on `run_id` / `browser_event_id`. The one-shot script cleans their existing dups, but they slowly re-accrue until guarded; I-16 is scoped to `observation` so it does not flap on them in the meantime. (AP-13 documents this.)


stevencarpenter:
Addressed all four automated review comments in 5789f61:

1. **opencode `compute_content_hash` (codex-connector, P2)** — now also hashes diff stats, diff text, and files touched (all rendered into `build_opencode_enrichment_prompt`). A file-only/diff-only change now perturbs the hash, so `decide_enqueue` re-enriches it instead of skipping (the gate previously caused a coverage miss). Test updated to assert diff stats + files touched change the hash while title/model (metadata) do not.
2. **`dedup-knowledge-nodes.py` orphan-vector risk (Copilot)** — `run()` now distinguishes `knowledge_vectors` *absent* (benign skip — fresh install / test fixture) from *present-but-unreachable* (vec0 not loaded), and aborts `--apply` in the latter so it can never delete nodes while orphaning their vectors. Added a regression test.
3. **`watchdog.rs` u64→i64 cast (Copilot)** — threshold now goes through `i64::try_from(...).unwrap_or(i64::MAX)`; no negative wrap on an absurd config value.
4. **`is_uuid_v4_stem` misnomer (Copilot)** — renamed to `is_uuid_stem` (it validates the 8-4-4-4-12 lowercase-hex shape, not the v4 version/variant).

Full Rust (fmt/clippy/test) and Python (937 pytest + ruff) suites green.

PR #211, opened 2026-06-08
fix(doctor): suppress source-staleness alarms for in-flight Codex/Cursor sessions still settling within min_idle_secs (v0.28.7)

stevencarpenter:


stevencarpenter:
## Addressed CodeRabbit review (fail-open for alerting)

Fixed in `eceeaa0`.

**CodeRabbit's finding (valid):** on config-load failure the `codex_session_state` / `cursor_session_state` closures returned `(false, false, false)`, which matched the `!*_sessions_exist =>` "no sessions found" suppression arm — silently hiding a genuinely stale `source_health` row behind a misleading reason. Both now return `(true, true, false)`, which falls through every suppression arm to `None` so a real ingestion failure still alarms. This matches the `*_enabled` fallback already used a few lines below (`unwrap_or((true, true, true))`).

**Extended beyond the flagged lines:** `opencode_db_recent` had the identical bug class (`return false` on config error → matched the "opencode DB idle" arm). Now returns `true` on config error for the same fail-open-for-alerting behavior. The shared comment block above `doctor_config` was updated to describe the new semantics.

A clean-context review confirmed all three fallbacks reach `None`, that the `*_enabled` flags read the same config binding (no split-brain), and that no other config-load fallback in `check_source_staleness` shares the bug. `cargo fmt`/`clippy`/298 lib tests pass.

